PR #1, opened 2022-02-26
Implement NAT traversal + Secure Connection

shutterbug2000:


jonbarrow:
I like the idea so far and this looks good, but can I see some usage examples to get a feel for how the end result would look?

jonbarrow:
(usage sent over Discord)

jonbarrow:
Some comments seem to have been left unchanged so they document the wrong functions

PR #7, opened 2023-04-23
Use PasswordFromPID function from NEX server

DaniElectra:
Since we have to get a password on the server for HPP, check that function to avoid confusion.


jonbarrow:
Looks good, just needs updated Go modules and it's ready

PR #10, opened 2023-05-21
Add AutoMatchmake_Postpone common method

shutterbug2000:
This gets MK7 into a match

### DIFF
--- a/globals/matchmaking_globals.go
+++ b/globals/matchmaking_globals.go
@@ -1,0 +1,12 @@
+package common_globals
+import (
+	match_making "github.com/PretendoNetwork/nex-protocols-go/match-making"
+)
+
+type CommonMatchmakeSession struct {
+	GameMatchmakeSession match_making.MatchmakeSession //used by the game, contains the current state of the MatchmakeSession
+	SearchMatchmakeSession match_making.MatchmakeSession //used by the server when searching for matches, contains the state of the MatchmakeSession during the search process for easy compares
+	PlayersByConnectionId []uint32 //players in the room, referenced by their connection IDs. This is used instead of the PID in order to ensure we're talking to the correct client (in case of e.g. multiple logins)
+}
+
+var Sessions = []CommonMatchmakeSession{}
--- a/globals/matchmaking_globals.go
+++ b/globals/matchmaking_globals.go
@@ -6,7 +6,7 @@ import (
 type CommonMatchmakeSession struct {
 	GameMatchmakeSession match_making.MatchmakeSession //used by the game, contains the current state of the MatchmakeSession
 	SearchMatchmakeSession match_making.MatchmakeSession //used by the server when searching for matches, contains the state of the MatchmakeSession during the search process for easy compares
-	PlayersByConnectionId []uint32 //players in the room, referenced by their connection IDs. This is used instead of the PID in order to ensure we're talking to the correct client (in case of e.g. multiple logins)
+	ConnectionIDs []uint32 //players in the room, referenced by their connection IDs. This is used instead of the PID in order to ensure we're talking to the correct client (in case of e.g. multiple logins)
 }
 
 var Sessions = []CommonMatchmakeSession{}
--- a/matchmake-extension/auto_matchmake_postpone.go
+++ b/matchmake-extension/auto_matchmake_postpone.go
@@ -1,0 +1,102 @@
+package matchmake_extension
+
+import (
+	"math"
+
+	"github.com/PretendoNetwork/mario-kart-7-secure/globals"
+	nex "github.com/PretendoNetwork/nex-go"
+	match_making "github.com/PretendoNetwork/nex-protocols-go/match-making"
+	matchmake_extension "github.com/PretendoNetwork/nex-protocols-go/matchmake-extension"
+	"github.com/PretendoNetwork/nex-protocols-go/notifications"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/globals"
+	"encoding/json"
+)
+
+func AutoMatchmake_Postpone(err error, client *nex.Client, callID uint32, matchmakeSession *match_making.MatchmakeSession, message string) {
+	missingHandler := false
+	if commonMatchmakeExtensionProtocol.CleanupSearchMatchmakeSessionHandler == nil {
+		logger.Warning("MatchmakeExtension::AutoMatchmake_Postpone missing CleanupSearchMatchmakeSessionHandler!")
+		missingHandler = true
+	}
+	if missingHandler {
+		return
+	}
+	//This is the best way I found to copy the full object. And I still hate it.
+	tmp, _ := json.Marshal(matchmakeSession)
+	matchmakeSessionCopy := match_making.NewMatchmakeSession()
+	json.Unmarshal(tmp, &matchmakeSessionCopy)
+	searchMatchmakeSession := commonMatchmakeExtensionProtocol.CleanupSearchMatchmakeSessionHandler(*matchmakeSessionCopy)
+	sessionIndex := FindSearchMatchmakeSession(searchMatchmakeSession)
+	if sessionIndex == math.MaxUint32 {
+		session := common_globals.CommonMatchmakeSession{
+			SearchMatchmakeSession: searchMatchmakeSession,
+			GameMatchmakeSession:   *matchmakeSession,
+		}
+		sessionIndex = len(common_globals.Sessions)
+		common_globals.Sessions = append(common_globals.Sessions, session)
+		common_globals.Sessions[sessionIndex].GameMatchmakeSession.Gathering.ID = uint32(sessionIndex)
+		common_globals.Sessions[sessionIndex].GameMatchmakeSession.Gathering.OwnerPID = client.PID()
+		common_globals.Sessions[sessionIndex].GameMatchmakeSession.Gathering.HostPID = client.PID()
+	}
+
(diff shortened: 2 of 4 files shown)


matchmake-extension/auto_matchmake_postpone.go
```diff
@@ -0,0 +1,102 @@
...
+	globals.NEXServer.Send(responsePacket)
+
+	rmcMessage := nex.NewRMCRequest()
+	rmcMessage.SetProtocolID(notifications.ProtocolID)
+	rmcMessage.SetCallID(0xffff0000 + callID)
+	rmcMessage.SetMethodID(notifications.MethodProcessNotificationEvent)
+
+	oEvent := notifications.NewNotificationEvent()
+	oEvent.PIDSource = common_globals.Sessions[sessionIndex].GameMatchmakeSession.Gathering.HostPID
+	oEvent.Type = 3001 // New participant
```
jonbarrow, line 82: Should we break these things out into something more readable? It would be nice to not have to rely on previous knowledge or other peoples comments to know what certain types and flags and such are in some structures like here

Not sure where that should go, though. Maybe as part of the notifications/nintendo-notifications protocols in nex-protocols-go? That way they could be accessed both here and in servers which might want to send notifications manually. Something like `notifications.NotificationTypes.NewMatchmakeParticipant` would be nice

globals/matchmaking_globals.go
```diff
@@ -0,0 +1,12 @@
+package common_globals
+import (
+	match_making "github.com/PretendoNetwork/nex-protocols-go/match-making"
+)
+
+type CommonMatchmakeSession struct {
+	GameMatchmakeSession match_making.MatchmakeSession //used by the game, contains the current state of the MatchmakeSession
+	SearchMatchmakeSession match_making.MatchmakeSession //used by the server when searching for matches, contains the state of the MatchmakeSession during the search process for easy compares
+	PlayersByConnectionId []uint32 //players in the room, referenced by their connection IDs. This is used instead of the PID in order to ensure we're talking to the correct client (in case of e.g. multiple logins)
```
jonbarrow: In Go the style guidelines dictate that identifiers like `ConnectionId` should actually be `ConnectionID` (this also applies to other names like `HTTP` and `URL`) https://github.com/golang/go/wiki/CodeReviewComments#initialisms

The name `PlayersByConnectionID`, while descriptive and correct, does feel a bit long-winded and awkward. Not exactly an issue and it can stay as is, since it is correct, but maybe something shorter like `UserIDs` or `ConnectionIDs` would be better?

Or alternatively, maybe we should store this based on the users PID, and then reference that client through the NEX server by the PID? I am unsure if the users connection ID would ever change but if it does then that would become an issue if we only store the connection ID. Again not an issue right now, just some food for thought. If it stays that's also fine :)

matchmake-extension/auto_matchmake_postpone.go
```diff
@@ -0,0 +1,102 @@
...
+
+func AutoMatchmake_Postpone(err error, client *nex.Client, callID uint32, matchmakeSession *match_making.MatchmakeSession, message string) {
+	missingHandler := false
+	if commonMatchmakeExtensionProtocol.CleanupSearchMatchmakeSessionHandler == nil {
+		logger.Warning("MatchmakeExtension::AutoMatchmake_Postpone missing CleanupSearchMatchmakeSessionHandler!")
+		missingHandler = true
+	}
+	if missingHandler {
+		return
+	}
```
jonbarrow: I see the benefit in storing `missingHandler` for when checking for multiple handlers, but in cases where there's only one check maybe we should just bail right away and not create any unnecessary variables?

matchmake-extension/auto_matchmake_postpone.go
```diff
@@ -0,0 +1,102 @@
...
+	//This is the best way I found to copy the full object. And I still hate it.
+	tmp, _ := json.Marshal(matchmakeSession)
+	matchmakeSessionCopy := match_making.NewMatchmakeSession()
+	json.Unmarshal(tmp, &matchmakeSessionCopy)
```
jonbarrow: Is this what you asked about on Discord for earlier? I think the best way forward in that case is with the example I gave earlier, by implementing a `Copy` method to `StructureInterface` and implementing that on each structure. We don't have to do this all at once, we can add this method incrementally as we have types which need to be copied

Because I agree, I do not like this at all, and this is likely much slower than just creating a new struct and moving data

matchmake-extension/auto_matchmake_postpone.go
```diff
@@ -0,0 +1,102 @@
...
+	common_globals.Sessions[sessionIndex].PlayersByConnectionId = append(common_globals.Sessions[sessionIndex].PlayersByConnectionId, client.ConnectionID())
+
+	rmcResponseStream := nex.NewStreamOut(globals.NEXServer)
+	rmcResponseStream.WriteString("MatchmakeSession")
+	lengthStream := nex.NewStreamOut(globals.NEXServer)
+	lengthStream.WriteStructure(&common_globals.Sessions[sessionIndex].GameMatchmakeSession)
+	matchmakeSessionLength := uint32(len(lengthStream.Bytes()))
+	rmcResponseStream.WriteUInt32LE(matchmakeSessionLength + 4)
+	rmcResponseStream.WriteUInt32LE(matchmakeSessionLength)
+	rmcResponseStream.WriteStructure(&common_globals.Sessions[sessionIndex].GameMatchmakeSession)
```
jonbarrow: What is happening here? This looks like you're manually writing the data for a DataHolder type? If so, `StreamOut` has a `WriteDataHolder` method made for this very task

Also when looking at Kinnay's wiki the DataHolder type is `Gathering` but you are writing a `MatchmakeSession`?

matchmake-extension/auto_matchmake_postpone.go
```diff
@@ -0,0 +1,102 @@
...
+	rmcResponseBody := rmcResponseStream.Bytes()
+
+	rmcResponse := nex.NewRMCResponse(matchmake_extension.ProtocolID, callID)
+	rmcResponse.SetSuccess(matchmake_extension.MethodAutoMatchmake_Postpone, rmcResponseBody)
+
+	rmcResponseBytes := rmcResponse.Bytes()
+
+	responsePacket, _ := nex.NewPacketV0(client, nil)
+
+	responsePacket.SetVersion(0)
```
jonbarrow: Always setting the packet version to 0 regardless of the servers version

matchmake-extension/auto_matchmake_postpone.go
```diff
@@ -0,0 +1,102 @@
...
+
+	stream := nex.NewStreamOut(globals.NEXServer)
+	oEventBytes := oEvent.Bytes(stream)
+	rmcMessage.SetParameters(oEventBytes)
+	rmcMessageBytes := rmcMessage.Bytes()
+
+	targetClient := globals.NEXServer.FindClientFromPID(uint32(common_globals.Sessions[sessionIndex].GameMatchmakeSession.Gathering.HostPID))
+
+	messagePacket, _ := nex.NewPacketV0(targetClient, nil)
+	messagePacket.SetVersion(1)
```
jonbarrow: Always setting the packet version to 1 regardless of the servers version

matchmake-extension/utility.go
```diff
@@ -0,0 +1,32 @@
+package matchmake_extension
+
+import (
+	//nex "github.com/PretendoNetwork/nex-go"
+	match_making "github.com/PretendoNetwork/nex-protocols-go/match-making"
+	//"github.com/PretendoNetwork/plogger-go"
+    "reflect"
```
jonbarrow: Maybe this is GitHubs code viewer playing tricks on me but is this line indented differently than the other stuff?

matchmake-extension/utility.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func FindSearchMatchmakeSession(searchMatchmakeSession match_making.MatchmakeSession) int {
+	returnSessionIndex := math.MaxUint32
+	//this portion finds any sessions that match the search session. It does not care about anything beyond that, such as if the match is already full. This is handled below.
+	candidateSessionIndexes := []int{}
+	for index, session := range common_globals.Sessions {
+		if reflect.DeepEqual(session.SearchMatchmakeSession, searchMatchmakeSession) {
+			candidateSessionIndexes = append(candidateSessionIndexes, index)
+		}
```
jonbarrow: Oh God do we not have a better way to do this besides using reflect? Reflect is super slow most of the time. I'm tempted to suggest adding an `Equals` API to `StructureInterface` much like `Copy` and checking for equal structs that way

matchmake-extension/utility.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+	for _, sessionIndex := range candidateSessionIndexes {
+		sessionToCheck := common_globals.Sessions[sessionIndex]
+		if(len(sessionToCheck.PlayersByConnectionId) >= int(sessionToCheck.GameMatchmakeSession.MaximumParticipants)){
+			continue
+		}else{
+			returnSessionIndex = sessionIndex //found a match
+			break
+		}
```
jonbarrow: Go does not require parentheses in `if` statements. Though it is valid syntax, it is not consistent with the rest of the codebase which does not use them, and the spacing used here both in `if` and `else` between the braces is not consistent with the rest of the codebase either (just me being picky about the style guidelines)

jonbarrow (changes requested):
Couple suggestions for changes and a few style comments

PR #11, opened 2023-05-22
Matchmaking protocol additions and changes

shutterbug2000:
- Move utility methods to global so any MatchMaking protocol can use them
- Add EndParticipation to Matchmaking Ext
- Remove clients from Sessions when they leave the server


matchmaking/protocol.go
```diff
@@ -55,5 +57,17 @@ func NewCommonMatchMakingProtocol(server *nex.Server) *CommonMatchMakingProtocol
...
+		server.On("Kick", func(packet *nex.PacketV0) {
+			fmt.Println("Leaving")
+			common_globals.RemoveConnectionIDFromAllSessions(packet.Sender().ConnectionID())
+		})
+	} else {
+		server.On("Kick", func(packet *nex.PacketV1) {
+			fmt.Println("Leaving")
+			common_globals.RemoveConnectionIDFromAllSessions(packet.Sender().ConnectionID())
+		})
+	}
```
DaniElectra:
```suggestion
	server.On("Kick", func(packet nex.PacketInterface) {
		fmt.Println("Leaving")
		common_globals.RemoveConnectionIDFromAllSessions(packet.Sender().ConnectionID())
	})
```

We can use PacketInterface here

globals/matchmaking_globals.go
```diff
@@ -9,4 +11,60 @@ type CommonMatchmakeSession struct {
...
+			Sessions[gathering].ConnectionIDs = DeleteIndex(Sessions[gathering].ConnectionIDs, index)
+		}
+	}
+	if len(Sessions[gathering].ConnectionIDs) == 0{
+		Sessions[gathering] = CommonMatchmakeSession{} //TODO: is there a better way to do this? right now, the index is used as the gathering id, and I'm not sure if there's a good way to remove without reordering.
+	}
+}
+
+func FindClientSessions(clientConnectionID uint32) []uint32 {
+	//TODO: is there any real instance where a client could be in multiple MatchmakeSessions at once? afaik there isn't, but I handle it regardless just in case (if we're sure there isn't, we could remove the handling and speed the code up)
```
DaniElectra: The only thing like this that appear on servers are Mario Kart 7 communities, but those use `PersistentGathering`. Then the communities have a list of active MatchmakeSessions, so I'm sure there's no way a client could be in multiple MatchmakeSessions

globals/matchmaking_globals.go
```diff
@@ -9,4 +11,60 @@ type CommonMatchmakeSession struct {
...
+	//this portion finds any sessions that match the search session. It does not care about anything beyond that, such as if the match is already full. This is handled below.
+	candidateSessionIndexes := []int{}
+	for index, session := range Sessions {
+		if reflect.DeepEqual(session.SearchMatchmakeSession, searchMatchmakeSession) { // TODO - for Jon: Equals in StructureInterface
+			candidateSessionIndexes = append(candidateSessionIndexes, index)
+		}
+	}
+	for _, sessionIndex := range candidateSessionIndexes {
+		sessionToCheck := Sessions[sessionIndex]
+		if len(sessionToCheck.ConnectionIDs) >= int(sessionToCheck.GameMatchmakeSession.MaximumParticipants){
```
DaniElectra:
```suggestion
		if len(sessionToCheck.ConnectionIDs) >= int(sessionToCheck.GameMatchmakeSession.MaximumParticipants) {
```
missing space

globals/matchmaking_globals.go
```diff
@@ -9,4 +11,60 @@ type CommonMatchmakeSession struct {
...
 
-var Sessions = []CommonMatchmakeSession{}
\ No newline at end of file
+var Sessions = []CommonMatchmakeSession{}
+
+func DeleteIndex(s []uint32, index int) []uint32 {
+    return append(s[:index], s[index+1:]...)
+}
+
+func RemoveConnectionIDFromRoom(clientConnectionID uint32, gathering uint32){
```
DaniElectra:
```suggestion
func RemoveConnectionIDFromRoom(clientConnectionID uint32, gathering uint32) {
```
missing space

globals/matchmaking_globals.go
```diff
@@ -9,4 +11,60 @@ type CommonMatchmakeSession struct {
...
+	for index, connectionID := range Sessions[gathering].ConnectionIDs {
+		if connectionID == clientConnectionID {
+			Sessions[gathering].ConnectionIDs = DeleteIndex(Sessions[gathering].ConnectionIDs, index)
+		}
+	}
+	if len(Sessions[gathering].ConnectionIDs) == 0{
```
jonbarrow:
```suggestion
	if len(Sessions[gathering].ConnectionIDs) == 0 {
```

missing space

globals/matchmaking_globals.go
```diff
@@ -9,4 +11,60 @@ type CommonMatchmakeSession struct {
...
+		Sessions[gathering] = CommonMatchmakeSession{} //TODO: is there a better way to do this? right now, the index is used as the gathering id, and I'm not sure if there's a good way to remove without reordering.
```
jonbarrow: If we are using a known key then using a `map` would be faster as the data grows larger. In Go when reading data from either type, a `map` is `O(1)` whereas a slice is `O(n)` based on the size of the slice

globals/matchmaking_globals.go
```diff
@@ -9,4 +11,60 @@ type CommonMatchmakeSession struct {
...
+	}
+}
+
+func FindClientSessions(clientConnectionID uint32) []uint32 {
+	//TODO: is there any real instance where a client could be in multiple MatchmakeSessions at once? afaik there isn't, but I handle it regardless just in case (if we're sure there isn't, we could remove the handling and speed the code up)
+	gatheringsFoundIn := []uint32{}
```
jonbarrow:
```suggestion
	gatheringsFoundIn := make([]uint32, 0, len(Sessions))
```

Creating a slice like this is only really useful when we don't know the length the slice will eventually be, and even then `make` can still be used in that case

In this case we know the max length the slice can be, since it can't be any larger than `len(Sessions)`. `make([]uint32, 0, len(Sessions))` will make a uint32 slice with 0 elements but has a max capacity of `len(Sessions)`. Makes it easier for Go to manage this memory

globals/matchmaking_globals.go
```diff
@@ -9,4 +11,60 @@ type CommonMatchmakeSession struct {
...
+		for _, connectionID := range Sessions[gatheringIndex].ConnectionIDs {
+			if connectionID == clientConnectionID {
+				gatheringsFoundIn = append(gatheringsFoundIn, uint32(gatheringIndex))
+			}
+		}
+	}
+	return gatheringsFoundIn
+}
+
+func RemoveConnectionIDFromAllSessions(clientConnectionID uint32){
```
jonbarrow:
```suggestion
func RemoveConnectionIDFromAllSessions(clientConnectionID uint32) {
```

Missing space

globals/matchmaking_globals.go
```diff
@@ -9,4 +11,60 @@ type CommonMatchmakeSession struct {
...
+	foundSessions := FindClientSessions(clientConnectionID)
+	for gatheringIndex, _ := range foundSessions {
+		RemoveConnectionIDFromRoom(clientConnectionID, uint32(gatheringIndex))
+	}
+}
+
+func FindSearchMatchmakeSession(searchMatchmakeSession match_making.MatchmakeSession) int {
+	returnSessionIndex := math.MaxUint32
+	//this portion finds any sessions that match the search session. It does not care about anything beyond that, such as if the match is already full. This is handled below.
+	candidateSessionIndexes := []int{}
```
jonbarrow: Same comment as before about using `make` here since we know the max capacity this slice can be

globals/matchmaking_globals.go
```diff
@@ -9,4 +11,60 @@ type CommonMatchmakeSession struct {
...
+	for index, session := range Sessions {
+		if reflect.DeepEqual(session.SearchMatchmakeSession, searchMatchmakeSession) { // TODO - for Jon: Equals in StructureInterface
+			candidateSessionIndexes = append(candidateSessionIndexes, index)
+		}
+	}
+	for _, sessionIndex := range candidateSessionIndexes {
+		sessionToCheck := Sessions[sessionIndex]
```
jonbarrow, line 60: Instead of looping to find indexes, and then also looping over the indexes later, why not just do this logic inside the first loop? Also line break between the `for` loops

jonbarrow (changes requested):

globals/matchmaking_globals.go
```diff
@@ -9,4 +11,60 @@ type CommonMatchmakeSession struct {
...
+				gatheringsFoundIn = append(gatheringsFoundIn, uint32(gatheringIndex))
+			}
+		}
+	}
+	return gatheringsFoundIn
+}
+
+func RemoveConnectionIDFromAllSessions(clientConnectionID uint32){
+	foundSessions := FindClientSessions(clientConnectionID)
+	for gatheringIndex, _ := range foundSessions {
```
jonbarrow: If you only need the index, a standard `for` loop is likely more performant since `for range` already reads the value at the index during iteration, which is thrown out here

globals/matchmaking_globals.go
```diff
@@ -9,4 +11,60 @@ type CommonMatchmakeSession struct {
...
+	}
+	if len(Sessions[gathering].ConnectionIDs) == 0{
+		Sessions[gathering] = CommonMatchmakeSession{} //TODO: is there a better way to do this? right now, the index is used as the gathering id, and I'm not sure if there's a good way to remove without reordering.
+	}
+}
+
+func FindClientSessions(clientConnectionID uint32) []uint32 {
+	//TODO: is there any real instance where a client could be in multiple MatchmakeSessions at once? afaik there isn't, but I handle it regardless just in case (if we're sure there isn't, we could remove the handling and speed the code up)
+	gatheringsFoundIn := []uint32{}
+	for gatheringIndex, _ := range Sessions {
```
jonbarrow: If you only need the index, a standard `for` loop is likely more performant since `for range` already reads the value at the index during iteration, which is thrown out here

jonbarrow (changes requested):

globals/matchmaking_globals.go
shutterbug2000: I did it this way in case we want to handle more complex matchmaking logic
e.g. instead of just filling the first room we find, search for one with more players, closer VR/whatever skill metric, etc.
so we can find rooms that we can use, and then narrow down the best one

globals/matchmaking_globals.go
shutterbug2000: As in 
`gatheringIndex := 0; gatheringIndex <= len(foundSessions); gatheringIndex++` ? or just removing the _?

jonbarrow:
```suggestion
	for i := 0; i < len(foundSessions); i++ {
```

(`i` can be changed that's just my usual default)

globals/matchmaking_globals.go
jonbarrow: Gotcha, sounds good then. Shouldn't be a huge performance issue regardless, was just looking for some micro-optimizations and ways to clean things up. But that makes sense

globals/matchmaking_globals.go
shutterbug2000: One last thing and I think this is ready: does this apply now that we're using a map for Sessions? since the keys aren't gonna be in order every time anymore

jonbarrow: No, when using a map type this is the right way to go. `for range` doesn't return an index, it returns a key (it just so happens that with a slice those are the same thing), so this is the right way to go to get these keys

jonbarrow: Though I would say that the name `gatheringIndex` is misleading since it's no longer an index

jonbarrow (approved):

DaniElectra (approved):

PR #12, opened 2023-06-05
Lots of changes on common protocols

DaniElectra:
- Since we now store the client station URLs on nex.Client, there's no
need for external handlers to do so, so remove them. Also, the handlers
were deprecated as they didn't do anything anyway.
- Use common struct on NAT traversal and remove external handlers,
as we now store the client station URLs on nex.Client. And implement
ReportNATTraversalResult.
- Various changes to Matchmaking:
  - Use structure.Copy() and structure.Equal() now that they are
  implemented.
  - Add notification types that weren't used before.
  - Remove handlers for MatchMaking(-Ext) and use global sessions for
  handling rooms.
  - And possibly more
  
Depends on https://github.com/PretendoNetwork/nex-protocols-go/pull/24

### DIFF
--- a/globals/matchmaking_globals.go
+++ b/globals/matchmaking_globals.go
@@ -5,8 +5,8 @@ import (
 )
 
 type CommonMatchmakeSession struct {
-	GameMatchmakeSession match_making.MatchmakeSession //used by the game, contains the current state of the MatchmakeSession
-	SearchMatchmakeSession match_making.MatchmakeSession //used by the server when searching for matches, contains the state of the MatchmakeSession during the search process for easy compares
+	GameMatchmakeSession *match_making.MatchmakeSession //used by the game, contains the current state of the MatchmakeSession
+	SearchMatchmakeSession *match_making.MatchmakeSession //used by the server when searching for matches, contains the state of the MatchmakeSession during the search process for easy compares
 	ConnectionIDs []uint32 //players in the room, referenced by their connection IDs. This is used instead of the PID in order to ensure we're talking to the correct client (in case of e.g. multiple logins)
 }
 
@@ -14,7 +14,20 @@ var Sessions map[uint32]*CommonMatchmakeSession
 var CurrentGatheringID uint32
 
 func DeleteIndex(s []uint32, index int) []uint32 {
-    return append(s[:index], s[index+1:]...)
+	if len(s) <= 1 {
+		return make([]uint32, 0)
+	}
+
+	return append(s[:index], s[index+1:]...)
+}
+
+func FindOtherConnectionID(myConnectionID uint32, gathering uint32) uint32 {
+	for _, connectionID := range Sessions[gathering].ConnectionIDs {
+		if connectionID != myConnectionID {
+			return connectionID
+		}
+	}
+	return math.MaxUint32
 }
 
 func RemoveConnectionIDFromRoom(clientConnectionID uint32, gathering uint32) {
@@ -41,17 +54,17 @@ func FindClientSession(clientConnectionID uint32) uint32 {
 
 func RemoveConnectionIDFromAllSessions(clientConnectionID uint32) {
 	foundSession := FindClientSession(clientConnectionID)
-	if(foundSession != math.MaxUint32){
+	if foundSession != math.MaxUint32 {
 		RemoveConnectionIDFromRoom(clientConnectionID, foundSession)
 	}
 }
 
-func FindSearchMatchmakeSession(searchMatchmakeSession match_making.MatchmakeSession) int {
+func FindSearchMatchmakeSession(searchMatchmakeSession *match_making.MatchmakeSession) int {
 	returnSessionIndex := math.MaxUint32
 	//this portion finds any sessions that match the search session. It does not care about anything beyond that, such as if the match is already full. This is handled below.
 	candidateSessionIndexes := make([]uint32, 0, len(Sessions))
 	for index, session := range Sessions {
-		if session.SearchMatchmakeSession.Equals(&searchMatchmakeSession) {
+		if session.SearchMatchmakeSession.Equals(searchMatchmakeSession) {
 			candidateSessionIndexes = append(candidateSessionIndexes, index)
 		}
 	}
--- a/globals/matchmaking_globals.go
+++ b/globals/matchmaking_globals.go
@@ -2,7 +2,6 @@ package common_globals
 import (
 	match_making "github.com/PretendoNetwork/nex-protocols-go/match-making"
 	"math"
-	"reflect"
 )
 
 type CommonMatchmakeSession struct {
@@ -25,7 +24,7 @@ func RemoveConnectionIDFromRoom(clientConnectionID uint32, gathering uint32) {
 		}
 	}
 	if len(Sessions[gathering].ConnectionIDs) == 0 {
-        delete(Sessions, gathering)
+		delete(Sessions, gathering)
 	}
 }
 
@@ -52,7 +51,7 @@ func FindSearchMatchmakeSession(searchMatchmakeSession match_making.MatchmakeSes
 	//this portion finds any sessions that match the search session. It does not care about anything beyond that, such as if the match is already full. This is handled below.
 	candidateSessionIndexes := make([]uint32, 0, len(Sessions))
 	for index, session := range Sessions {
-		if reflect.DeepEqual(session.SearchMatchmakeSession, searchMatchmakeSession) { // TODO - for Jon: Equals in StructureInterface
+		if session.SearchMatchmakeSession.Equals(&searchMatchmakeSession) {
 			candidateSessionIndexes = append(candidateSessionIndexes, index)
 		}
 	}
--- a/globals/matchmaking_globals.go
+++ b/globals/matchmaking_globals.go
@@ -1,35 +1,50 @@
 package common_globals
 import (
-	match_making "github.com/PretendoNetwork/nex-protocols-go/match-making"
 	"math"
+
+	match_making "github.com/PretendoNetwork/nex-protocols-go/match-making"
 )
 
 type CommonMatchmakeSession struct {
-	GameMatchmakeSession *match_making.MatchmakeSession //used by the game, contains the current state of the MatchmakeSession
-	SearchMatchmakeSession *match_making.MatchmakeSession //used by the server when searching for matches, contains the state of the MatchmakeSession during the search process for easy compares
-	ConnectionIDs []uint32 //players in the room, referenced by their connection IDs. This is used instead of the PID in order to ensure we're talking to the correct client (in case of e.g. multiple logins)
+	GameMatchmakeSession *match_making.MatchmakeSession // Used by the game, contains the current state of the MatchmakeSession
+	SearchMatchmakeSession *match_making.MatchmakeSession // Used by the server when searching for matches, contains the state of the MatchmakeSession during the search process for easy compares
+	ConnectionIDs []uint32 // Players in the room, referenced by their connection IDs. This is used instead of the PID in order to ensure we're talking to the correct client (in case of e.g. multiple logins)
 }
 
 var Sessions map[uint32]*CommonMatchmakeSession
-var CurrentGatheringID uint32
 
-func DeleteIndex(s []uint32, index int) []uint32 {
-	if len(s) <= 1 {
-		return make([]uint32, 0)
+// GetSessionIndex returns a gathering ID which doesn't belong to any session
+func GetSessionIndex() uint32 {
+	var gatheringID uint32 = 1
+	for gatheringID < math.MaxUint32 {
+		// If the session does not exist, the gathering ID is empty and can be used
+		if _, ok := Sessions[gatheringID]; !ok {
+			return gatheringID
+		}
+
+		gatheringID++
 	}
 
-	return append(s[:index], s[index+1:]...)
+	return 0
 }
 
+// DeleteIndex removes a value from a slice with the given index
+func DeleteIndex(s []uint32, index int) []uint32 {
+	s[index] = s[len(s)-1]
+	return s[:len(s)-1]
+}
+
+// FindOtherConnectionID searches a connection ID on the gathering that isn't the given one
 func FindOtherConnectionID(myConnectionID uint32, gathering uint32) uint32 {
 	for _, connectionID := range Sessions[gathering].ConnectionIDs {
 		if connectionID != myConnectionID {
 			return connectionID
 		}
 	}
-	return math.MaxUint32
+	return 0
 }
 
+// RemoveConnectionIDFromRoom removes a client from the gathering
 func RemoveConnectionIDFromRoom(clientConnectionID uint32, gathering uint32) {
 	for index, connectionID := range Sessions[gathering].ConnectionIDs {
 		if connectionID == clientConnectionID {
@@ -41,6 +56,7 @@ func RemoveConnectionIDFromRoom(clientConnectionID uint32, gathering uint32) {
 	}
 }
 
+// FindClientSession searches the gathering where the client is on
 func FindClientSession(clientConnectionID uint32) uint32 {
 	for gatheringID := range Sessions {
 		for _, connectionID := range Sessions[gatheringID].ConnectionIDs {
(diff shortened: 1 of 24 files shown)


matchmake-extension/auto_matchmake_postpone.go
```diff
@@ -34,10 +30,16 @@ func AutoMatchmake_Postpone(err error, client *nex.Client, callID uint32, matchm
 		common_globals.Sessions[sessionIndex].GameMatchmakeSession.Gathering.ID = uint32(sessionIndex)
 		common_globals.Sessions[sessionIndex].GameMatchmakeSession.Gathering.OwnerPID = client.PID()
 		common_globals.Sessions[sessionIndex].GameMatchmakeSession.Gathering.HostPID = client.PID()
+
+		currentTime := nex.NewDateTime(0)
+		currentTime = nex.NewDateTime(currentTime.UTC())
```
jonbarrow: This is being immediately reassigned?

matchmaking-ext/end_participation.go
```diff
@@ -4,10 +4,19 @@ import (
...
 	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/globals"
+	"github.com/PretendoNetwork/nex-protocols-go/notifications"
 )
 
 func EndParticipation(err error, client *nex.Client, callID uint32, idGathering uint32, strMessage string) {
 	server := commonMatchMakingExtProtocol.server
+	matchmakeSession := common_globals.Sessions[idGathering].GameMatchmakeSession
+	ownerPID := matchmakeSession.Gathering.OwnerPID
+	if client.PID() == matchmakeSession.Gathering.OwnerPID {
+		if matchmakeSession.Gathering.Flags & 0x10 == 0 {
```
jonbarrow: What is this flag `0x10`, and do we know of any more flags? Having raw flag/type values in code like this is often unclear as to what the intention actually is

nat-traversal/report_nat_traversal_result.go
```diff
@@ -2,35 +2,35 @@ package nattraversal
...
 		responsePacket.SetVersion(1)
-		responsePacket.SetSource(0xA1)
-		responsePacket.SetDestination(0xAF)
-		responsePacket.SetType(nex.DataPacket)
-		responsePacket.SetPayload(rmcResponseBytes)
+	}
 
-		responsePacket.AddFlag(nex.FlagNeedsAck)
-		responsePacket.AddFlag(nex.FlagReliable)
+	responsePacket.SetVersion(1)
```
jonbarrow: Packet version always being set to 1

globals/matchmaking_globals.go
```diff
@@ -52,7 +51,7 @@ func FindSearchMatchmakeSession(searchMatchmakeSession match_making.MatchmakeSes
 	//this portion finds any sessions that match the search session. It does not care about anything beyond that, such as if the match is already full. This is handled below.
 	candidateSessionIndexes := make([]uint32, 0, len(Sessions))
 	for index, session := range Sessions {
-		if reflect.DeepEqual(session.SearchMatchmakeSession, searchMatchmakeSession) { // TODO - for Jon: Equals in StructureInterface
+		if session.SearchMatchmakeSession.Equals(&searchMatchmakeSession) {
```
jonbarrow: I see a lot of referencing and dereferencing going on all throughout here, is there a reason for that? We usually just stick with passing pointers around, mostly because it's faster with larger data structures (and we have very large and complex data structures at times)

jonbarrow (changes requested):

matchmaking-ext/end_participation.go
DaniElectra: I took the flag from the docs: https://nintendo-wiki.pretendo.network/docs/nex/protocols/match-making/types#flags
I will add a comment noting this

There are other flags like 0x20 and 0x200, but they are unknown

globals/matchmaking_globals.go
DaniElectra: The Session globals store the `MatchmakeSession` without pointers, so I adapted the code to what was done.

I will change it to pointers

jonbarrow: Gotcha, I must not have caught that in a previous commit. That's my bad, thank you

DaniElectra:
While I was testing this on MK7, everything seemed to work fine, except that one of the clients would randomly disconnect in-game from the session (also causing the other client to leave) and from NEX for seemingly no reason (there aren't any requests sent to NEX before this happening, just the `Leaving` message).

Luckily, the NEX matchmaking sessions themselves seem to work fine. I can't test nat traversal however, since both machines are connected to the same network

jonbarrow:
> (there aren't any requests sent to NEX before this happening, just the `Leaving` message).

About how long does this take to happen? It sounds like a ping timeout issue

DaniElectra:
> > (there aren't any requests sent to NEX before this happening, just the `Leaving` message).
> 
> About how long does this take to happen? It sounds like a ping timeout issue

It would happen some seconds after the game had started (the time varied on each test).
Actually, that may be it, since the server is set with ping after [999 seconds](https://github.com/PretendoNetwork/mario-kart-7-secure/blob/70a61fa40a6a0159d97deb5bafc09d7ac540fb7a/nex/server.go#L28) and I hadn't changed that parameter

I will try again tomorrow

DaniElectra:
Interestingly, the issue above appears on Race mode, but on Battle mode *almost* everything works just fine, and I can play as many rounds as I want. However, when one of the participants leaves, the `EndParticipation` isn't sent for some reason (but the client doesn't disconnect from the server), and the other player will show a connection error.

I wonder if this is a *me* issue, because of having the server and consoles on the same network? idk

jonbarrow:
It's possible that's causing issues, yeah. I've had a few issues when hosting the servers on the same network as the clients in the past. That is weird though

jonbarrow:
I see this was marked as open, no longer a draft. Are we still waiting for more changes or is this ready for review now?

DaniElectra:
Yes, it's ready for review

globals/matchmaking_globals.go
```diff
@@ -2,20 +2,32 @@ package common_globals
...
 }
 
 var Sessions map[uint32]*CommonMatchmakeSession
 var CurrentGatheringID uint32
 
 func DeleteIndex(s []uint32, index int) []uint32 {
-    return append(s[:index], s[index+1:]...)
+	if len(s) <= 1 {
+		return make([]uint32, 0)
+	}
```
jonbarrow: Is there a reason why a new slice is allocated in these cases?

globals/matchmaking_globals.go
```diff
@@ -2,20 +2,32 @@ package common_globals
...
+
+	return append(s[:index], s[index+1:]...)
```
jonbarrow: Does the order of the slice matter when we are removing an index? If not, then there are some alternative solutions to this. Using `append` like this is called reslicing in Go and is pretty expensive. [See here](https://stackoverflow.com/a/37335777) for some more information about the speed plus an alternative solution (though comments do point out some things to be careful of)

If order does matter, or if we need to remain with the `append` method for some other reason, then we can replace this function with `slices.Delete`, an experimental feature in Go 1.18 using generics. [Docs here](https://pkg.go.dev/golang.org/x/exp/slices#Delete) and [source code here](https://cs.opensource.google/go/x/exp/+/0b5c67f0:slices/slices.go;l=156). `slices.Delete` is just a wrapper around the `append` method used here. Despite the documentation saying `it does not create a new slice`, however, if used you MUST still reassign the slice to update the length

globals/matchmaking_globals.go
```diff
@@ -2,20 +2,32 @@ package common_globals
...
+}
+
+func FindOtherConnectionID(myConnectionID uint32, gathering uint32) uint32 {
```
jonbarrow, line 38: Only going to comment this once in case it occurs elsewhere in the code, but some of these function names aren't very clear. `FindOtherConnectionID` is pretty vague, does it mean finding another users connection ID? Is it finding your own connection ID in another gathering? Something else?

If function names themselves cannot be improved on (whether that be because they are generic by design, the name becomes too long, etc) then we should add some comments detailing what it's used for. Really we should be adding Godoc comments to all exported types/functions/etc anyway in all of our codebases

globals/matchmaking_globals.go
```diff
@@ -42,17 +54,17 @@ func FindClientSession(clientConnectionID uint32) uint32 {
...
 func RemoveConnectionIDFromAllSessions(clientConnectionID uint32) {
 	foundSession := FindClientSession(clientConnectionID)
-	if(foundSession != math.MaxUint32){
+	if foundSession != math.MaxUint32 {
 		RemoveConnectionIDFromRoom(clientConnectionID, foundSession)
 	}
 }
 
-func FindSearchMatchmakeSession(searchMatchmakeSession match_making.MatchmakeSession) int {
+func FindSearchMatchmakeSession(searchMatchmakeSession *match_making.MatchmakeSession) int {
```
jonbarrow: This is another function with a weird/vague name. It implies it will will return a MatchmakeSession data type, but in reality it returns an index into `Sessions`

globals/matchmaking_globals.go
```diff
@@ -42,17 +54,17 @@ func FindClientSession(clientConnectionID uint32) uint32 {
...
 	returnSessionIndex := math.MaxUint32
```
jonbarrow: Although very unlikely (nearing impossible), is IS _technically possible_ for this value to be hit. In cases where an index either could not be found or is invalid, a truly invalid index should be returned, such as -1

matchmake-extension/auto_matchmake_postpone.go
```diff
@@ -8,41 +8,46 @@ import (
...
+			GameMatchmakeSession:   matchmakeSession,
 		}
 		sessionIndex = common_globals.CurrentGatheringID
 		common_globals.Sessions[sessionIndex] = &session
 		common_globals.Sessions[sessionIndex].GameMatchmakeSession.Gathering.ID = uint32(sessionIndex)
 		common_globals.Sessions[sessionIndex].GameMatchmakeSession.Gathering.OwnerPID = client.PID()
 		common_globals.Sessions[sessionIndex].GameMatchmakeSession.Gathering.HostPID = client.PID()
+
+		currentTime := nex.NewDateTime(0)
+		common_globals.Sessions[sessionIndex].GameMatchmakeSession.StartedTime = nex.NewDateTime(currentTime.UTC())
```
jonbarrow: You don't have to create a new `DateTime` here. `DateTime.UTC()` does return a new `DateTime`, but it also updates the current instances value

matchmake-extension/auto_matchmake_postpone.go
```diff
@@ -8,41 +8,46 @@ import (
...
+
 		common_globals.CurrentGatheringID++
```
jonbarrow: This value doesn't appear to be checked anywhere. Though very unlikely, if a server runs for long enough and people keep creating sessions then this will reach the max uint32. Reaching this value may create unexpected side effects (such as the above check thinking no session was found, which I addressed in an earlier comment as well). This should be bounds checked

matchmaking-ext/end_participation.go
jonbarrow: Gonna open this back up again just to avoid making another comment on the same line. I know flags `0x20` and `0x200` are unknown, but is there a good name for flag `0x10`? ~~Like maybe `KeepSessionAlive` or `DontDeleteSession`?~~ after looking at the following `else` block, maybe a better name would be `ChangeOwner`. I'm thinking of adding these flags to `nex-protocols-go` similar to the notification types

matchmaking-ext/end_participation.go
```diff
@@ -40,4 +62,110 @@ func EndParticipation(err error, client *nex.Client, callID uint32, idGathering
...
+	oEvent := notifications.NewNotificationEvent()
+	oEvent.PIDSource = otherClient.PID()
+	oEvent.Type = notifications.NotificationTypes.OwnershipChanged
+	oEvent.Param1 = gathering
+	oEvent.Param2 = otherClient.PID()
+
+	// TODO - This doesn't seem to appear on all servers
+	// https://github.com/kinnay/NintendoClients/issues/101
+	// unixTime := time.Now()
+	// oEvent.StrParam = strconv.FormatInt(unixTime.UnixMicro(), 10)
```
jonbarrow: This comment implies that the string parameter is not always present. This is not what the linked issue says, the linked issue states that a 3rd uint32 parameter is not always present . And indeed this is true, this 3rd uint32 parameter was added in NEX 3.5 https://nintendo-wiki.pretendo.network/docs/nex/protocols/notifications

This string parameter should always be set, and the 3rd uint32 should only be sent in NEX 3.5+

matchmaking/update_session_host.go
```diff
@@ -46,35 +39,35 @@ func updateSessionHost(err error, client *nex.Client, callID uint32, gid uint32)
...
+	oEvent.Type = notifications.NotificationTypes.OwnershipChanged
 	oEvent.Param1 = gid
 	oEvent.Param2 = client.PID()
 
+	// TODO - This doesn't seem to appear on all servers
 	// https://github.com/kinnay/NintendoClients/issues/101
-	unixTime := time.Now()
-	oEvent.StrParam = strconv.FormatInt(unixTime.UnixMicro(), 10)
+	// unixTime := time.Now()
+	// oEvent.StrParam = strconv.FormatInt(unixTime.UnixMicro(), 10)
```
jonbarrow: Same comment as before with regards to this string parameter

jonbarrow (changes requested):
Some things addressed here are not actually changes you made, but they are around the code you are working on and this PR seems to be broadly covering a refactor of the common protocols, so I figured they fit

Also full transparency, since I am personally not very familiar with the matchmaking system specifically, my review is strictly on code quality and performance. I have not exactly reviewed the logic behind the matchmaking itself, and I would prefer if I could get another set of eyes on that. I have already assigned @shutterbug2000 for this task as well, but will likely also call upon Rambo since he's done great work on MK8

You will also need to update the go modules, as I have pushed updates to them and made some breaking changes to some protocols

globals/matchmaking_globals.go
DaniElectra: If the slice only has 1 value, using `return append(s[:index], s[index+1:]...)` would make an out of bounds check `s[1+1]`, so I just returned an empty slice. I will remove this and use an alternative solution from below, however.

matchmake-extension/auto_matchmake_postpone.go
DaniElectra: `DateTime.UTC()` returns a [uint64](https://github.com/PretendoNetwork/nex-go/blob/81594ad4892d8163f7754798586212cda65ff66e/types.go#L341). But you are right, it also updates the current instance

matchmaking-ext/end_participation.go
DaniElectra: I guess the comment wasn't clear, my bad. It means that other servers don't store a Unix time value on the `StrParam`, those servers leave the `StrParam` empty.
I will edit it to make it more clear

matchmaking-ext/end_participation.go
DaniElectra: Yeah, I agree. Maybe something like `DisconnectChangeOwner`, since it happens when the original owner disconnects?

shutterbug2000:
Looks good to me

jonbarrow:
Relating to a comment we talked about the other day, `nex-protocols-go` version 1.0.30 now has a `GatheringFlags` structure in the match making protocol to hold all the possible flags a gathering can have. I added `0x20` and `0x200` as well despite them being labeled as "unknown"

jonbarrow:
Carrying over from Discord, @shutterbug2000 has confirmed that the matchmaking changes here work fine in MK7 during his testing. So this PR should be good to merge once the Go modules are updated (I merged your other PR https://github.com/PretendoNetwork/nex-protocols-go/pull/25 and made a new release) and once the other issues are looked at

DaniElectra:
I changed invalid values to be zero, as it can't be a valid gathering ID nor a valid connection ID. To get a gathering ID, I changed the counter to the function `GetSessionIndex()`, which will check for empty sessions starting from ID 1. I tested this implementation in isolation (with maps with a million values) with minimal performance impact. Considering a PID can only create one session, it's impossible that we reach the max uint32 limit, so if we get an error, it should be an issue of not erasing empty sessions.

I also implemented time validation on Kerberos tickets. If the ticket is expired, we kick the client for timeout, as on Mario Kart 7 the client was still sending data packets when using a graceful kick (using graceful kicks aren't really an issue though, as we delete the client and the server can't decrypt those data packets anymore `[PRUDPv0] Error decoding packet data: [PRUDPv0] Error parsing RMC request: RMC Request size does not match length of buffer`, but the client would get spammy with those packets).

jonbarrow (approved):

PR #13, opened 2023-06-20
Set  RVConnectionData time as DateTime

DaniElectra:
~~Set the structure version to 1 if NEX version is 3.5+~~

### DIFF
--- a/authentication/login.go
+++ b/authentication/login.go
@@ -45,6 +45,12 @@ func login(err error, client *nex.Client, callID uint32, username string) {
 		serverTime := nex.NewDateTime(0)
 		pConnectionData.SetTime(serverTime.UTC())
 
+		nexVersion := commonAuthenticationProtocol.server.NEXVersion()
+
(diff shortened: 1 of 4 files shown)


jonbarrow:
Why is this being done in this library? This feels like something `nex-go` should handle when encoding the data in the stream

DaniElectra:
We can't set the version on `nex-go` because by the time we get to `RVConnectionData.Bytes()` the structure version is already written in the stream. Unless we set the structure version when creating the structure?

DaniElectra:
Actually no, my bad. I was thinking that the version was written before calling `Structure.Bytes()`

~~Should I close this PR or wait for updating the modules?~~ I actually have to update the time value for `RVConnectionData` anyway

jonbarrow:
Is this good to close then?

DaniElectra:
No, I will use this PR to update the time for `RVConnectionData` from uint64 to a DateTime

*sorry about this confusion lmao*

PR #14, opened 2023-07-12
Implement methods for Pokemon Gen6

shutterbug2000:
This allows Wonder Trade and Random Matchup free battle to function. PSS matchup will require more methods.


globals/matchmaking_globals.go
```diff
@@ -9,6 +9,7 @@ import (
 type CommonMatchmakeSession struct {
 	GameMatchmakeSession   *match_making_types.MatchmakeSession // Used by the game, contains the current state of the MatchmakeSession
 	SearchMatchmakeSession *match_making_types.MatchmakeSession // Used by the server when searching for matches, contains the state of the MatchmakeSession during the search process for easy compares
+	SearchCriteria		   []*match_making_types.MatchmakeSessionSearchCriteria // Used by the server when searching for matches, contains the list of MatchmakeSessionSearchCriteria
```
DaniElectra: You are mixing tabs and spaces between `SearchCriteria` and `[]*match_making_types.MatchmakeSessionSearchCriteria`. Please use spaces

globals/matchmaking_globals.go
```diff
@@ -99,3 +100,30 @@ func SearchGatheringWithMatchmakeSession(searchMatchmakeSession *match_making_ty
...
 }
+
+// SearchGatheringWithMatchmakeSession finds a gathering that matches with a MatchmakeSession
+func SearchGatheringWithSearchCriteria(lstSearchCriteria []*match_making_types.MatchmakeSessionSearchCriteria) uint32 {
+	var returnSessionIndex uint32 = 0
+
+	// This portion finds any sessions that match the search session. It does not care about anything beyond that, such as if the match is already full. This is handled below.
+	candidateSessionIndexes := make([]uint32, 0, len(Sessions))
+	for index, session := range Sessions {
+		if(len(lstSearchCriteria) == len(session.SearchCriteria)){
```
DaniElectra:
```suggestion
		if len(lstSearchCriteria) == len(session.SearchCriteria) {
```

globals/matchmaking_globals.go
```diff
@@ -99,3 +100,30 @@ func SearchGatheringWithMatchmakeSession(searchMatchmakeSession *match_making_ty
 	}
 	return returnSessionIndex
 }
+
+// SearchGatheringWithMatchmakeSession finds a gathering that matches with a MatchmakeSession
```
DaniElectra: The comment mismatches the function name `SearchGatheringWithSearchCriteria`

matchmake-extension/auto_matchmake_with_search_criteria_postpone.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+
+		common_globals.Sessions[sessionIndex] = &session
+		common_globals.Sessions[sessionIndex].GameMatchmakeSession.Gathering.ID = sessionIndex
+		common_globals.Sessions[sessionIndex].GameMatchmakeSession.Gathering.OwnerPID = client.PID()
+		common_globals.Sessions[sessionIndex].GameMatchmakeSession.Gathering.HostPID = client.PID()
+
+		common_globals.Sessions[sessionIndex].GameMatchmakeSession.StartedTime = nex.NewDateTime(0)
+		common_globals.Sessions[sessionIndex].GameMatchmakeSession.StartedTime.UTC()
+		matchmakeSession.SessionKey = []byte("00000000000000000000000000000000")
```
DaniElectra:
```suggestion
		matchmakeSession.SessionKey = make([]byte, 32)
```

matchmaking-ext/end_participation.go
```diff
@@ -11,7 +11,35 @@ import (
 
 func endParticipation(err error, client *nex.Client, callID uint32, idGathering uint32, strMessage string) {
 	server := commonMatchMakingExtProtocol.server
-	matchmakeSession := common_globals.Sessions[idGathering].GameMatchmakeSession
+	session, ok := common_globals.Sessions[idGathering]
+	if(!ok){
```
DaniElectra:
```suggestion
	if !ok {
```

globals/matchmaking_globals.go
```diff
@@ -99,3 +100,30 @@ func SearchGatheringWithMatchmakeSession(searchMatchmakeSession *match_making_ty
...
+	}
+	for _, sessionIndex := range candidateSessionIndexes {
+		sessionToCheck := Sessions[sessionIndex]
+		if len(sessionToCheck.ConnectionIDs) >= int(sessionToCheck.GameMatchmakeSession.MaximumParticipants) {
+			continue
+		} else {
+			returnSessionIndex = sessionIndex //found a match
+			break
+		}
+	}
```
DaniElectra: We also have to skip the sessions who have `OpenParticipation` disabled. This applies here and on `SearchGatheringWithMatchmakeSession`.

```suggestion
	for _, sessionIndex := range candidateSessionIndexes {
		sessionToCheck := Sessions[sessionIndex]
		if len(sessionToCheck.ConnectionIDs) >= int(sessionToCheck.GameMatchmakeSession.MaximumParticipants) {
			continue
		}
		
		if !sessionToCheck.GameMatchmakeExtension.OpenParticipation {
			continue
		}
		
		returnSessionIndex = sessionIndex // Found a match
		break
	}
```

This didn't affect on Mario Kart 7 before because the field was set to `true` by default.

matchmake-extension/auto_matchmake_with_search_criteria_postpone.go
```diff
@@ -0,0 +1,126 @@
...
+	// so let's make sure the client is removed from the session
+	common_globals.RemoveConnectionIDFromAllSessions(client.ConnectionID())
+
+	var matchmakeSession *match_making_types.MatchmakeSession
+	anyGatheringDataType := anyGathering.TypeName()
+
+	if anyGatheringDataType == "MatchmakeSession" {
+		matchmakeSession = anyGathering.ObjectData().(*match_making_types.MatchmakeSession)
+	}
+
```
DaniElectra, line 32: Should we also cleanup the search criteria parameters here just in case?

matchmake-extension/open_participation.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+func openParticipation(err error, client *nex.Client, callID uint32, gid uint32) {
+	server := commonMatchmakeExtensionProtocol.server
+	session, ok := common_globals.Sessions[gid]
+	rmcResponse := nex.NewRMCResponse(matchmake_extension.ProtocolID, callID)
+	if ok {
+		session.GameMatchmakeSession.OpenParticipation = true
+		rmcResponse.SetSuccess(matchmake_extension.MethodOpenParticipation, nil)
+	}else{
```
DaniElectra:
```suggestion
	} else {
```

nat-traversal/get_relay_signature_key.go
```diff
@@ -0,0 +1,45 @@
...
+	rmcResponseStream := nex.NewStreamOut(server)
+
+	rmcResponseStream.WriteInt32LE(0)
+	dateTime := nex.NewDateTime(0)
+	dateTime.Now()
+	rmcResponseStream.WriteDateTime(dateTime)
+	rmcResponseStream.WriteString("")
+	rmcResponseStream.WriteUInt16LE(0)
+	rmcResponseStream.WriteInt32LE(0)
+	rmcResponseStream.WriteUInt32LE(0x55d01)
```
DaniElectra: We know the contents of some of this fields, we don't need to hardcode it

```suggestion
	rmcResponseStream.WriteInt32LE(0) // relayMode
	dateTime := nex.NewDateTime(0)
	dateTime.UTC()
	rmcResponseStream.WriteDateTime(dateTime) // currentUTCTime
	rmcResponseStream.WriteString("") // address
	rmcResponseStream.WriteUInt16LE(0) // port
	rmcResponseStream.WriteInt32LE(0) // relayAddressType
	rmcResponseStream.WriteUInt32LE() // gameServerID. Allow the implementation to set this field with a handler (something like `SetGameServerID()`?)
```

Also, are the address and port *really* empty? Aren't they the address and port of the client or something?

matchmake-extension/auto_matchmake_with_search_criteria_postpone.go
```diff
@@ -0,0 +1,126 @@
...
+	rmcMessage.SetProtocolID(notifications.ProtocolID)
+	rmcMessage.SetCallID(0xffff0000 + callID)
+	rmcMessage.SetMethodID(notifications.MethodProcessNotificationEvent)
+
+	category := notifications.NotificationCategories.Participation
+	subtype := notifications.NotificationSubTypes.Participation.NewParticipant
+
+	oEvent := notifications_types.NewNotificationEvent()
+	oEvent.PIDSource = common_globals.Sessions[sessionIndex].GameMatchmakeSession.Gathering.HostPID
+	oEvent.Type = notifications.BuildNotificationType(category, subtype)
```
DaniElectra, line 104: This function isn't available in the current go modules of the package. Please update them.

Also please update the rest of notifications to use the new format

DaniElectra (changes requested):
Some various comments :p

matchmake-extension/auto_matchmake_with_search_criteria_postpone.go
jonbarrow: Is this supposed to be all null bytes? I thought these were supposed to be randomized?

matchmake-extension/auto_matchmake_with_search_criteria_postpone.go
jonbarrow: I was going to mention this in my PR as well but instead of duplicating the comment I'll just add this here:

If this type is ever not `MatchmakeSession` then the `matchmakeSession` variable will remain in its zero state and will have nil pointers. I assume that because you're checking the type name that means there's a chance it could be a different type, those cases should probably be handled (even if it's an `else` statement here for now that logs a warning and bails or something)

nat-traversal/get_relay_signature_key.go
shutterbug2000: It's supposed to be the address of the relay server. We don't have an implementation of the relay server, so the response tells the client to not use it.

matchmake-extension/open_participation.go
```diff
@@ -0,0 +1,40 @@
...
+	nex "github.com/PretendoNetwork/nex-go"
+	matchmake_extension "github.com/PretendoNetwork/nex-protocols-go/matchmake-extension"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/globals"
+)
+
+func openParticipation(err error, client *nex.Client, callID uint32, gid uint32) {
+	server := commonMatchmakeExtensionProtocol.server
+	session, ok := common_globals.Sessions[gid]
+	rmcResponse := nex.NewRMCResponse(matchmake_extension.ProtocolID, callID)
+	if ok {
```
jonbarrow: The more idiomatic way of writing these statments in Go is

```suggestion
	if session, ok := common_globals.Sessions[gid]; ok {
```

matchmaking-ext/end_participation.go
jonbarrow: Same comment as before for me. The more idiomatic way of writing these statments in Go is

```suggestion
	if session, ok := common_globals.Sessions[idGathering]; !ok {
```

jonbarrow (changes requested):

globals/matchmaking_globals.go
```diff
@@ -99,3 +100,34 @@ func SearchGatheringWithMatchmakeSession(searchMatchmakeSession *match_making_ty
...
+		if len(sessionToCheck.ConnectionIDs) >= int(sessionToCheck.GameMatchmakeSession.MaximumParticipants) {
+			continue
+		}
+		
+		if !sessionToCheck.GameMatchmakeExtension.OpenParticipation {
+			continue
+		}
+		
+		returnSessionIndex = sessionIndex // Found a match
+		break
```
jonbarrow: This assign and break just immediately invokes the following `return returnSessionIndex` anyway

```suggestion
		return sessionIndex // Found a match
```

matchmake-extension/auto_matchmake_with_search_criteria_postpone.go
shutterbug2000: Yeah, probably a wise idea.

matchmaking-ext/end_participation.go
```diff
@@ -11,7 +11,35 @@ import (
 
 func endParticipation(err error, client *nex.Client, callID uint32, idGathering uint32, strMessage string) {
 	server := commonMatchMakingExtProtocol.server
-	matchmakeSession := common_globals.Sessions[idGathering].GameMatchmakeSession
+	session, ok := common_globals.Sessions[idGathering]
+	if !ok {
```
jonbarrow: The more idiomatic way to write this in Go would be

```suggestion
	if session, ok := common_globals.Sessions[idGathering]; !ok {
```

jonbarrow (changes requested):

nat-traversal/get_relay_signature_key.go
jonbarrow: How much information do we have about the relay servers? Should I add this to the todo list?

matchmaking-ext/end_participation.go
shutterbug2000: This doesn't appear to assign the session var outside of the if statement. I could put the rest in an else{}, that should work, up to you

DaniElectra: You have to set the following right before if statement:

```go
var session *common_globals.CommonMatchmakeSession
```

jonbarrow: Yes that's true, when done this way the variable is scoped to the following block. I would suggest then defining the variables inside the scope they are supposed to be used in, then. Like so

```suggestion
	var session *common_globals.CommonMatchmakeSession
	var ok bool
	if session, ok = common_globals.Sessions[idGathering]; !ok {
```

nat-traversal/get_relay_signature_key.go
shutterbug2000: Honestly, not much info. I wouldn't make them a priority for now.

matchmake-extension/auto_matchmake_postpone.go
```diff
@@ -25,6 +25,9 @@ func autoMatchmake_Postpone(err error, client *nex.Client, callID uint32, anyGat
 
 	if anyGatheringDataType == "MatchmakeSession" {
 		matchmakeSession = anyGathering.ObjectData().(*match_making_types.MatchmakeSession)
+	}else{
```
jonbarrow:
```suggestion
	} else {
```

matchmake-extension/auto_matchmake_with_search_criteria_postpone.go
```diff
@@ -0,0 +1,134 @@
...
+	// A client may disconnect from a session without leaving reliably,
+	// so let's make sure the client is removed from the session
+	common_globals.RemoveConnectionIDFromAllSessions(client.ConnectionID())
+
+	var matchmakeSession *match_making_types.MatchmakeSession
+	anyGatheringDataType := anyGathering.TypeName()
+
+	if anyGatheringDataType == "MatchmakeSession" {
+		matchmakeSession = anyGathering.ObjectData().(*match_making_types.MatchmakeSession)
+	}else{
```
jonbarrow:
```suggestion
	} else {
```

matchmake-extension/open_participation.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	nex "github.com/PretendoNetwork/nex-go"
+	matchmake_extension "github.com/PretendoNetwork/nex-protocols-go/matchmake-extension"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/globals"
+)
+
+func openParticipation(err error, client *nex.Client, callID uint32, gid uint32) {
+	server := commonMatchmakeExtensionProtocol.server
+	rmcResponse := nex.NewRMCResponse(matchmake_extension.ProtocolID, callID)
+	if session, ok := common_globals.Sessions[gid]; ok{
```
jonbarrow:
```suggestion
	if session, ok := common_globals.Sessions[gid]; ok {
```

jonbarrow (changes requested):
Just some spacing changes. ~~After this the only thing left is the Go modules here need to be updated as @DaniElectra mentioned, and then~~ (EDIT: Never mind I was mistaken about the library version being used here) this looks good to go!

jonbarrow (approved):

jonbarrow:
Looks good to me

PR #15, opened 2023-07-27
Update to new naming style in nex-protocols-go

SuperMarioDaBom:
As title says; confirmed does compile and run

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,8 +3,8 @@ module github.com/PretendoNetwork/nex-protocols-common-go
 go 1.19
 
 require (
-	github.com/PretendoNetwork/nex-go v1.0.28
-	github.com/PretendoNetwork/nex-protocols-go v1.0.39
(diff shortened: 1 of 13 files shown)


go.mod
```diff
@@ -3,8 +3,8 @@ module github.com/PretendoNetwork/nex-protocols-common-go
 go 1.19
 
 require (
-	github.com/PretendoNetwork/nex-go v1.0.28
-	github.com/PretendoNetwork/nex-protocols-go v1.0.39
+	github.com/PretendoNetwork/nex-go v1.0.34
+	github.com/PretendoNetwork/nex-protocols-go v1.0.44
```
jonbarrow: The new release is version 1.0.45, now 44. Looks like this was done before Go's cache updated

jonbarrow: I deleted this why does GitHub do this

SuperMarioDaBom: fun times

PR #17, opened 2023-08-31
Handle multiple station URLs on Register and more

DaniElectra:
- Games like Mario Kart 7 or Mario Tennis Open set the local station and
the public station. In these cases, use the public station given by the client.
- Add the PID and RVCID on both stations, as they are required on NAT
traversal and not all games will call `ReportNATProperties` to add them.
- Implement `MatchmakeExtension::CloseParticipation`
- and more

### DIFF
--- a/globals/matchmaking_globals.go
+++ b/globals/matchmaking_globals.go
@@ -12,3 +12,4 @@ type CommonMatchmakeSession struct {
 }
 
 var Sessions map[uint32]*CommonMatchmakeSession
+var CurrentGatheringID uint32
--- a/globals/matchmaking_globals.go
+++ b/globals/matchmaking_globals.go
@@ -1,6 +1,7 @@
 package common_globals
 
 import (
+	"github.com/PretendoNetwork/nex-go"
 	match_making_types "github.com/PretendoNetwork/nex-protocols-go/match-making/types"
 )
 
@@ -12,4 +13,4 @@ type CommonMatchmakeSession struct {
 }
 
 var Sessions map[uint32]*CommonMatchmakeSession
-var CurrentGatheringID uint32
+var CurrentGatheringID = nex.NewCounter(0)
--- a/secure-connection/register.go
+++ b/secure-connection/register.go
@@ -19,10 +19,16 @@ func register(err error, client *nex.Client, callID uint32, stationUrls []*nex.S
 	localStation := stationUrls[0]
 
 	// * A NEX client can set the public station URL by setting two URLs on the array
+	// * Check each URL for a public station
 	var publicStation *nex.StationURL
-	if len(stationUrls) > 1 {
-		publicStation = stationUrls[1]
-	} else {
+	for _, stationURL := range stationUrls {
+		if stationURL.Type() == 3 {
+			publicStation = stationURL
+			break
+		}
+	}
+
+	if publicStation == nil {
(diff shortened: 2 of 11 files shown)


globals/matchmaking_globals.go
```diff
@@ -12,3 +12,4 @@ type CommonMatchmakeSession struct {
 }
 
 var Sessions map[uint32]*CommonMatchmakeSession
+var CurrentGatheringID uint32
```
jonbarrow: Not a massive deal but `nex-go` already provides a `Counter` struct which is used internally there for things like client connection IDs and packet sequence IDs, but can also just be reused here

secure-connection/register.go
```diff
@@ -17,15 +17,27 @@ func register(err error, client *nex.Client, callID uint32, stationUrls []*nex.S
...
-	publicStation.SetAddress(client.Address().IP.String())
-	publicStation.SetPort(uint32(client.Address().Port))
-	publicStation.SetNatf(0)
-	publicStation.SetNatm(0)
-	publicStation.SetType(3)
+	// * A NEX client can set the public station URL by setting two URLs on the array
+	var publicStation *nex.StationURL
+	if len(stationUrls) > 1 {
+		publicStation = stationUrls[1]
+	} else {
```
jonbarrow: Public stations seem to use a `type` parameter of 3. This feels like a safer way to check for the public station rather than assuming it's the 2nd value in the list

Out of curiosity, how does it _know_ it's public station here?

secure-connection/register.go
```diff
@@ -17,15 +17,27 @@ func register(err error, client *nex.Client, callID uint32, stationUrls []*nex.S
...
+		publicStation.SetNatm(0)
+		publicStation.SetType(3)
+	}
+
+	localStation.SetPID(client.PID())
 	publicStation.SetPID(client.PID())
 
+	localStation.SetRVCID(client.ConnectionID())
+	publicStation.SetRVCID(client.ConnectionID())
+
```
jonbarrow, line 46: Has this been tested on a wide number of games? I tried always setting these values (PID and connection ID) on all the stations in `GetSessionURLs` while working on Minecraft, and it seemed to break PIA in Mario Kart 8. It could have been a fluke, but I tried a few times and the guest couldn't connect to the host until I reverted those changes

jonbarrow (changes requested):

secure-connection/register.go
DaniElectra: > Out of curiosity, how does it know it's public station here?

I'm not sure tbh. It seems to be a thing external to the server though, because it also uses the public IP even when the server is on the same network as the client

secure-connection/register.go
DaniElectra: > I tried always setting these values (PID and connection ID) on all the stations in GetSessionURLs while working on Minecraft, and it seemed to break PIA in Mario Kart 8.

That's weird. AFAIK, the PID and connection ID should be set on the stations anyway when calling `ReportNATProperties`. We can't rely on that for NEX 2.x games though, because they don't seem to call that method at any time, so I'm setting them on `Register` aswell.

jonbarrow: I'll merge this for now, and check Mario Kart 8 again. If it breaks, then we have to find a different solution

jonbarrow (approved):

PR #18, opened 2023-10-03
Implement various functions and player disconnects

DaniElectra:
This PR implements the `UpdateSessionURL` and the `JoinMatchmakeSession` functions, as well as better handling when a player leaves unreliably (after a timeout).

This implements the same functionality as on CTGP-7 Network, with the new addition of `JoinMatchmakeSession` (which I just copy-pasted from `JoinMatchmakeSessionWithParam` with little modifications) to allow players to join to their friends or people who they have played before.


jonbarrow:
Looks good to me :+1:

PR #19, opened 2023-10-07
Add common Ranking protocol

shutterbug2000:
Tested on both Ultimate NES Remix and DKC:TF without issues

### DIFF
--- a/ranking/get_cached_top_x_ranking.go
+++ b/ranking/get_cached_top_x_ranking.go
@@ -11,44 +11,43 @@ func getCachedTopXRanking(err error, client *nex.Client, callID uint32, category
 		logger.Warning("Ranking::GetCachedTopXRanking missing GetRankingsAndCountByCategoryAndRankingOrderParamHandler!")
 		return nex.Errors.Core.NotImplemented
 	}
+
 	rmcResponse := nex.NewRMCResponse(ranking.ProtocolID, callID)
 	server := client.Server()
 
 	if err != nil {
 		logger.Error(err.Error())
-		rmcResponse.SetError(nex.Errors.Ranking.Unknown)
+		return nex.Errors.Ranking.InvalidArgument
 	}
 
 	err, rankDataList, totalCount := commonRankingProtocol.getRankingsAndCountByCategoryAndRankingOrderParamHandler(category, orderParam)
 	if err != nil {
-		logger.Error(err.Error())
-		rmcResponse.SetError(nex.Errors.Ranking.Unknown)
+		logger.Critical(err.Error())
+		return nex.Errors.Ranking.Unknown
 	}
 
 	if totalCount == 0 || len(rankDataList) == 0 {
-		rmcResponse.SetError(nex.Errors.Ranking.NotFound)
+		return nex.Errors.Ranking.NotFound
 	}
 
-	if err == nil && totalCount != 0 {
-		rankingResult := ranking_types.NewRankingResult()
+	rankingResult := ranking_types.NewRankingResult()
 
-		rankingResult.RankDataList = rankDataList
-		rankingResult.TotalCount = totalCount
-		rankingResult.SinceTime = nex.NewDateTime(0x1f40420000) // * 2000-01-01T00:00:00.000Z, this is what the real server sends back
+	rankingResult.RankDataList = rankDataList
+	rankingResult.TotalCount = totalCount
+	rankingResult.SinceTime = nex.NewDateTime(0x1f40420000) // * 2000-01-01T00:00:00.000Z, this is what the real server sends back
 		
-		pResult := ranking_types.NewRankingCachedResult()
-		pResult.CreatedTime = nex.NewDateTime(0x1f40420000) //TODO: does this matter?
-		pResult.ExpiredTime = nex.NewDateTime(0x1f40420000) //TODO: does this matter?
-		pResult.MaxLength = 0
+	pResult := ranking_types.NewRankingCachedResult()
+	pResult.CreatedTime = nex.NewDateTime(nex.NewDateTime(serverTime.UTC()))
+	pResult.ExpiredTime = nex.NewDateTime(serverTime.FromTimestamp(time.Now().UTC()+300)) //The real server sends the "CreatedTime" + 5 minutes. It doesn't change, even on subsequent requests, until after the ExpiredTime has passed (seemingly what the "cached" means). Whether we need to replicate this idk, but in case, here's a note.
+	pResult.MaxLength = 10 //This is the length Ultimate NES Remix uses. TODO: Does this matter? and are other games different?
 
-		rmcResponseStream := nex.NewStreamOut(server)
+	rmcResponseStream := nex.NewStreamOut(server)
 
-		rmcResponseStream.WriteStructure(pResult)
+	rmcResponseStream.WriteStructure(pResult)
 
-		rmcResponseBody := rmcResponseStream.Bytes()
+	rmcResponseBody := rmcResponseStream.Bytes()
 
-		rmcResponse.SetSuccess(ranking.MethodGetCachedTopXRanking, rmcResponseBody)
-	}
+	rmcResponse.SetSuccess(ranking.MethodGetCachedTopXRanking, rmcResponseBody)
 
 	rmcResponseBytes := rmcResponse.Bytes()
 
--- a/ranking/get_cached_top_x_ranking.go
+++ b/ranking/get_cached_top_x_ranking.go
@@ -40,7 +40,9 @@ func getCachedTopXRanking(err error, client *nex.Client, callID uint32, category
 	pResult := ranking_types.NewRankingCachedResult()
 	serverTime := nex.NewDateTime(0)
 	pResult.CreatedTime = nex.NewDateTime(serverTime.UTC())
-	//The real server sends the "CreatedTime" + 5 minutes. It doesn't change, even on subsequent requests, until after the ExpiredTime has passed (seemingly what the "cached" means). Whether we need to replicate this idk, but in case, here's a note.
+	//The real server sends the "CreatedTime" + 5 minutes.
+	//It doesn't change, even on subsequent requests, until after the ExpiredTime has passed (seemingly what the "cached" means).
+	//Whether we need to replicate this idk, but in case, here's a note.
 	pResult.ExpiredTime = nex.NewDateTime(serverTime.FromTimestamp(time.Now().UTC().Add(time.Minute * time.Duration(5))))
 	pResult.MaxLength = 10 //This is the length Ultimate NES Remix uses. TODO: Does this matter? and are other games different?
 
--- a/ranking/get_cached_top_x_ranking.go
+++ b/ranking/get_cached_top_x_ranking.go
@@ -4,6 +4,8 @@ import (
 	"github.com/PretendoNetwork/nex-go"
 	ranking "github.com/PretendoNetwork/nex-protocols-go/ranking"
 	ranking_types "github.com/PretendoNetwork/nex-protocols-go/ranking/types"
+
+	"time"
 )
 
 func getCachedTopXRanking(err error, client *nex.Client, callID uint32, category uint32, orderParam *ranking_types.RankingOrderParam) uint32 {
@@ -20,7 +22,7 @@ func getCachedTopXRanking(err error, client *nex.Client, callID uint32, category
 		return nex.Errors.Ranking.InvalidArgument
 	}
 
-	err, rankDataList, totalCount := commonRankingProtocol.getRankingsAndCountByCategoryAndRankingOrderParamHandler(category, orderParam)
+	rankDataList, totalCount, err := commonRankingProtocol.getRankingsAndCountByCategoryAndRankingOrderParamHandler(category, orderParam)
 	if err != nil {
 		logger.Critical(err.Error())
 		return nex.Errors.Ranking.Unknown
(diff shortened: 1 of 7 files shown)


ranking/get_cached_top_x_ranking.go
```diff
@@ -0,0 +1,76 @@
...
+	if commonRankingProtocol.getRankingsAndCountByCategoryAndRankingOrderParamHandler == nil {
+		logger.Warning("Ranking::GetCachedTopXRanking missing GetRankingsAndCountByCategoryAndRankingOrderParamHandler!")
+		return nex.Errors.Core.NotImplemented
+	}
+	rmcResponse := nex.NewRMCResponse(ranking.ProtocolID, callID)
+	server := client.Server()
+
+	if err != nil {
+		logger.Error(err.Error())
+		rmcResponse.SetError(nex.Errors.Ranking.Unknown)
```
DaniElectra: Change all of these with returns please. This also reduces nesting for error checks.
Also, can you change the error returned from the first error check to `Ranking.InvalidArgument`?

```suggestion
		return nex.Errors.Ranking.InvalidArgument
```

ranking/protocol.go
```diff
@@ -0,0 +1,89 @@
...
+
+var commonRankingProtocol *CommonRankingProtocol
+var logger = plogger.NewLogger()
+
+type CommonRankingProtocol struct {
+	server             *nex.Server
+	DefaultProtocol    *ranking.Protocol
+	MarioKart8Protocol *ranking_mario_kart_8.Protocol
+
+	getCommonDataHandler                                     func(unique_id uint64) (error, []byte)
```
DaniElectra: Reorder these so that the error is the last thing that is returned

```suggestion
	getCommonDataHandler                                     func(unique_id uint64) ([]byte, error)
```

ranking/get_cached_top_x_ranking.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	err, rankDataList, totalCount := commonRankingProtocol.getRankingsAndCountByCategoryAndRankingOrderParamHandler(category, orderParam)
+	if err != nil {
+		logger.Error(err.Error())
```
DaniElectra: we are using Critical for database errors

```suggestion
		logger.Critical(err.Error())
```

ranking/get_ranking.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/PretendoNetwork/nex-go"
+	ranking "github.com/PretendoNetwork/nex-protocols-go/ranking"
+	ranking_types "github.com/PretendoNetwork/nex-protocols-go/ranking/types"
+)
+
+func getRanking(err error, client *nex.Client, callID uint32, rankingMode uint8, category uint32, orderParam *ranking_types.RankingOrderParam, uniqueID uint64, principalID uint32) uint32 {
+	if commonRankingProtocol.getRankingsAndCountByCategoryAndRankingOrderParamHandler == nil {
+		logger.Warning("Ranking::GetRanking missing GetRankingsAndCountByCategoryAndRankingOrderParamHandler!")
+		return nex.Errors.Core.NotImplemented
+	}
```
DaniElectra, line 13: add a space at the end

```suggestion
	if commonRankingProtocol.getRankingsAndCountByCategoryAndRankingOrderParamHandler == nil {
		logger.Warning("Ranking::GetRanking missing GetRankingsAndCountByCategoryAndRankingOrderParamHandler!")
		return nex.Errors.Core.NotImplemented
	}

```

ranking/get_cached_top_x_rankings.go
```diff
@@ -0,0 +1,78 @@
...
+			rankingResult := ranking_types.NewRankingResult()
+
+			rankingResult.RankDataList = rankDataList
+			rankingResult.TotalCount = totalCount
+			rankingResult.SinceTime = nex.NewDateTime(0x1f40420000) // * 2000-01-01T00:00:00.000Z, this is what the real server sends back
+			
+			result := ranking_types.NewRankingCachedResult()
+			result.CreatedTime = nex.NewDateTime(0x1f40420000) //TODO: does this matter?
+			result.ExpiredTime = nex.NewDateTime(0x1f40420000) //TODO: does this matter?
+			result.MaxLength = 0
```
DaniElectra: Is this actually zero?

ranking/get_cached_top_x_rankings.go
```diff
@@ -0,0 +1,78 @@
...
+		if err == nil && totalCount != 0 {
+			rankingResult := ranking_types.NewRankingResult()
+
+			rankingResult.RankDataList = rankDataList
+			rankingResult.TotalCount = totalCount
+			rankingResult.SinceTime = nex.NewDateTime(0x1f40420000) // * 2000-01-01T00:00:00.000Z, this is what the real server sends back
+			
+			result := ranking_types.NewRankingCachedResult()
+			result.CreatedTime = nex.NewDateTime(0x1f40420000) //TODO: does this matter?
+			result.ExpiredTime = nex.NewDateTime(0x1f40420000) //TODO: does this matter?
```
DaniElectra: We should set the expire time to some date in the far future like Dec 31st 9999, just in case

DaniElectra (changes requested):

ranking/get_common_data.go
```diff
@@ -0,0 +1,56 @@
...
+
+	if commonDataErr == nil {
+		rmcResponseStream := nex.NewStreamOut(server)
+
+		rmcResponseStream.WriteBuffer(commonData)
+
+		rmcResponseBody := rmcResponseStream.Bytes()
+
+		rmcResponse.SetSuccess(ranking.MethodGetCommonData, rmcResponseBody)
+	}
```
jonbarrow: I've noticed a couple places like this where `err` is only being checked if it *is* `nil`, and doing a success. But there is nothing being done for if it *isn't* `nil`

This should probably throw `Ranking::NotFound` if `commonDataErr != nil`. That way we follow "happy path", where the default flow is a success and these kinds of checks would be for errors and to bail early

ranking/get_ranking.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		logger.Error(err.Error())
+		rmcResponse.SetError(nex.Errors.Ranking.Unknown)
+	}
+
+	if totalCount == 0 || len(rankDataList) == 0 {
+		rmcResponse.SetError(nex.Errors.Ranking.NotFound)
+	}
+
+	if err == nil && totalCount != 0 {
```
jonbarrow: With Dani's suggestion to change lines 25 and 29 into returns, this check would no longer be needed so long as the check on line 28 is split into 2 checks

jonbarrow (changes requested):
Dani got most of the stuff I had to say already

ranking/get_cached_top_x_rankings.go
shutterbug2000: Left a comment in the code re: this. I'm not sure if it matters, but it's noted down either way :p

ranking/get_cached_top_x_rankings.go
shutterbug2000: Left a comment in the code re: this. I'm not sure if it matters, but it's noted down either way :p

ranking/get_ranking.go
shutterbug2000: I'm a bit confused by "so long as the check on line 28 is split into 2 checks" :p

ranking/get_cached_top_x_ranking.go
```diff
@@ -0,0 +1,78 @@
...
+	rankingResult := ranking_types.NewRankingResult()
+
+	rankingResult.RankDataList = rankDataList
+	rankingResult.TotalCount = totalCount
+	rankingResult.SinceTime = nex.NewDateTime(0x1f40420000) // * 2000-01-01T00:00:00.000Z, this is what the real server sends back
+		
+	pResult := ranking_types.NewRankingCachedResult()
+	serverTime := nex.NewDateTime(0)
+	pResult.CreatedTime = nex.NewDateTime(serverTime.UTC())
+	pResult.ExpiredTime = nex.NewDateTime(serverTime.FromTimestamp(time.Now().UTC().Add(time.Minute * time.Duration(5)))) //The real server sends the "CreatedTime" + 5 minutes. It doesn't change, even on subsequent requests, until after the ExpiredTime has passed (seemingly what the "cached" means). Whether we need to replicate this idk, but in case, here's a note.
```
DaniElectra: This comment is very long to be on a single line

```suggestion
	// The real server sends the "CreatedTime" + 5 minutes. It doesn't change, even on subsequent requests, until after the ExpiredTime has passed (seemingly what the "cached" means). Whether we need to replicate this idk, but in case, here's a note
	pResult.ExpiredTime = nex.NewDateTime(serverTime.FromTimestamp(time.Now().UTC().Add(time.Minute * time.Duration(5))))
```

ranking/upload_common_data.go
```diff
@@ -0,0 +1,53 @@
...
+	ranking "github.com/PretendoNetwork/nex-protocols-go/ranking"
+)
+
+func uploadCommonData(err error, client *nex.Client, callID uint32, commonData []byte, uniqueID uint64) uint32 {
+	if commonRankingProtocol.uploadCommonDataHandler == nil {
+		logger.Warning("Ranking::UploadCommonData missing UploadCommonDataHandler!")
+		return nex.Errors.Core.NotImplemented
+	}
+	
+	rmcResponse := nex.NewRMCResponse(ranking.ProtocolID, callID)
```
DaniElectra: The RMC response creation can be moved further down to be exactly before the `SetSuccess` function

ranking/protocol.go
```diff
@@ -0,0 +1,89 @@
+package ranking
+
+import (
+	"strings"
+
+	"github.com/PretendoNetwork/nex-go"
+	_ "github.com/PretendoNetwork/nex-protocols-go"
```
DaniElectra: This is imported on some common protocols so that we can access the `DataHolder` structures. There aren't any of them for the Ranking protocol, so this can be removed

ranking/get_common_data.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		logger.Error(err.Error())
+		return nex.Errors.Ranking.InvalidArgument
+	}
+
+	commonData, commonDataErr := commonRankingProtocol.getCommonDataHandler(uniqueID)
+
+	if commonDataErr != nil {
+		return nex.Errors.Ranking.NotFound
+	}
```
DaniElectra: little nitpick, the errors can all be renamed to `err` now that we are handling them with returns

```suggestion
	commonData, err := commonRankingProtocol.getCommonDataHandler(uniqueID)
	if err != nil {
		return nex.Errors.Ranking.NotFound
	}
```

ranking/get_common_data.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		logger.Error(err.Error())
+		return nex.Errors.Ranking.InvalidArgument
+	}
+
+	commonData, err := commonRankingProtocol.getCommonDataHandler(uniqueID)
+
+	if err != nil {
+		return nex.Errors.Ranking.NotFound
+	}
```
DaniElectra: The empty space can be removed

```suggestion
	commonData, err := commonRankingProtocol.getCommonDataHandler(uniqueID)
	if err != nil {
		return nex.Errors.Ranking.NotFound
	}
```

ranking/get_cached_top_x_ranking.go
```diff
@@ -0,0 +1,79 @@
...
+	rankingResult := ranking_types.NewRankingResult()
+
+	rankingResult.RankDataList = rankDataList
+	rankingResult.TotalCount = totalCount
+	rankingResult.SinceTime = nex.NewDateTime(0x1f40420000) // * 2000-01-01T00:00:00.000Z, this is what the real server sends back
+		
+	pResult := ranking_types.NewRankingCachedResult()
+	serverTime := nex.NewDateTime(0)
+	pResult.CreatedTime = nex.NewDateTime(serverTime.UTC())
+	//The real server sends the "CreatedTime" + 5 minutes. It doesn't change, even on subsequent requests, until after the ExpiredTime has passed (seemingly what the "cached" means). Whether we need to replicate this idk, but in case, here's a note.
```
DaniElectra: I would split the comment on multiple lines for better readability

ranking/get_cached_top_x_rankings.go
```diff
@@ -0,0 +1,83 @@
...
+		rankingResult := ranking_types.NewRankingResult()
+
+		rankingResult.RankDataList = rankDataList
+		rankingResult.TotalCount = totalCount
+		rankingResult.SinceTime = nex.NewDateTime(0x1f40420000) // * 2000-01-01T00:00:00.000Z, this is what the real server sends back
+			
+		result := ranking_types.NewRankingCachedResult()
+		serverTime := nex.NewDateTime(0)
+		result.CreatedTime = nex.NewDateTime(serverTime.UTC())
+		//The real server sends the "CreatedTime" + 5 minutes. It doesn't change, even on subsequent requests, until after the ExpiredTime has passed (seemingly what the "cached" means). Whether we need to replicate this idk, but in case, here's a note.
```
DaniElectra: I would split the comment on multiple lines for better readability

DaniElectra (approved):
some minor format changes, otherwise looks good. I will wait for Jon's final review before merging

jonbarrow:
LGTM :+1:

jonbarrow (approved):

PR #21, opened 2023-10-12
Move session creation and notification sending

shutterbug2000:
This helps reduce duplicate code and bugs that result from it.
Marked as a Draft for now since it needs testing with lots more games to ensure no regressions happen.

Tested with:
Minecraft: Wii U Edition
Tri-Force Heroes

Notable games to test still (these worked before this change):
Splatoon
Mario Kart 8
Mario Kart 7
Kid Icarus: Uprising

### DIFF
--- a/globals/matchmaking_globals.go
+++ b/globals/matchmaking_globals.go
@@ -14,4 +14,4 @@ type CommonMatchmakeSession struct {
 
 var Sessions map[uint32]*CommonMatchmakeSession
 var CurrentGatheringID = nex.NewCounter(0)
-var CurrentMatchmakingCallID uint32
+var CurrentMatchmakingCallID = nex.NewCounter(0)
--- a/globals/matchmaking_globals.go
+++ b/globals/matchmaking_globals.go
@@ -14,3 +14,4 @@ type CommonMatchmakeSession struct {
 
 var Sessions map[uint32]*CommonMatchmakeSession
 var CurrentGatheringID = nex.NewCounter(0)
+var CurrentMatchmakingCallID uint32
--- a/globals/matchmaking_utils.go
+++ b/globals/matchmaking_utils.go
@@ -5,7 +5,7 @@ import (
 	"math"
 	"strconv"
 	"strings"
-	"math/rand" 
+	"crypto/rand" 
 
 	nex "github.com/PretendoNetwork/nex-go"
 	match_making "github.com/PretendoNetwork/nex-protocols-go/match-making"
@@ -481,8 +481,6 @@ func AddPlayersToSession(session *CommonMatchmakeSession, connectionIDs []uint32
 		notificationRequestMessage.SetParameters(notificationStream.Bytes())
 		notificationRequestBytes := notificationRequestMessage.Bytes()
 
-		target := server.FindClientFromPID(uint32(session.GameMatchmakeSession.Gathering.OwnerPID))
-
 		var messagePacket nex.PacketInterface
 
 		if server.PRUDPVersion() == 0 {
@@ -502,6 +500,13 @@ func AddPlayersToSession(session *CommonMatchmakeSession, connectionIDs []uint32
 		messagePacket.AddFlag(nex.FlagReliable)
 
 		server.Send(messagePacket)
+
+		target := server.FindClientFromPID(uint32(session.GameMatchmakeSession.Gathering.OwnerPID))
+		if target == nil {
+			// TODO - Error here?
+			//logger.Warning("Player not found")
+			return nil, 0
+		}
 
 		if server.PRUDPVersion() == 0 {
 			messagePacket, _ = nex.NewPacketV0(target, nil)
--- a/globals/matchmaking_utils.go
+++ b/globals/matchmaking_utils.go
@@ -255,7 +255,7 @@ func FindSessionsByMatchmakeSessionSearchCriterias(lstSearchCriteria []*match_ma
 							continue
 						}
 
-						if session.SearchMatchmakeSession.MinimumParticipants < uint16(min) {
+						if session.GameMatchmakeSession.MinimumParticipants < uint16(min) {
 							continue
 						}
 					}
@@ -267,7 +267,7 @@ func FindSessionsByMatchmakeSessionSearchCriterias(lstSearchCriteria []*match_ma
 							continue
 						}
 
-						if session.SearchMatchmakeSession.MinimumParticipants > uint16(max) {
+						if session.GameMatchmakeSession.MinimumParticipants > uint16(max) {
 							continue
 						}
 					}
@@ -289,7 +289,7 @@ func FindSessionsByMatchmakeSessionSearchCriterias(lstSearchCriteria []*match_ma
 							continue
 						}
 
-						if session.SearchMatchmakeSession.MaximumParticipants < uint16(min) {
+						if session.GameMatchmakeSession.MaximumParticipants < uint16(min) {
 							continue
 						}
 					}
@@ -301,7 +301,7 @@ func FindSessionsByMatchmakeSessionSearchCriterias(lstSearchCriteria []*match_ma
 							continue
 						}
 
-						if session.SearchMatchmakeSession.MaximumParticipants > uint16(max) {
+						if session.GameMatchmakeSession.MaximumParticipants > uint16(max) {
 							continue
 						}
 					}
--- a/globals/matchmaking_utils.go
+++ b/globals/matchmaking_utils.go
@@ -5,7 +5,7 @@ import (
 	"math"
 	"strconv"
 	"strings"
-	"crypto/rand" 
+	"crypto/rand"
 
 	nex "github.com/PretendoNetwork/nex-go"
 	match_making "github.com/PretendoNetwork/nex-protocols-go/match-making"
--- a/globals/matchmaking_utils.go
+++ b/globals/matchmaking_utils.go
@@ -91,8 +91,7 @@ func RemoveClientFromAllSessions(client *nex.Client) {
 
 			rmcMessage := nex.NewRMCRequest()
 			rmcMessage.SetProtocolID(notifications.ProtocolID)
-			rmcMessage.SetCallID(CurrentMatchmakingCallID)
-			CurrentMatchmakingCallID+=1
+			rmcMessage.SetCallID(CurrentMatchmakingCallID.Increment())
 			rmcMessage.SetMethodID(notifications.MethodProcessNotificationEvent)
 
 			category := notifications.NotificationCategories.Participation
@@ -145,10 +144,9 @@ func RemoveClientFromAllSessions(client *nex.Client) {
 // CreateSessionByMatchmakeSession creates a gathering from a MatchmakeSession
 func CreateSessionByMatchmakeSession(matchmakeSession *match_making_types.MatchmakeSession, searchMatchmakeSession *match_making_types.MatchmakeSession, hostPID uint32) (*CommonMatchmakeSession, error, uint32) {
 	sessionIndex := GetAvailableGatheringID()
-	// This should in theory be impossible, as there aren't enough PIDs creating sessions to fill the uint32 limit.
-	// If we ever get here, we must be not deleting sessions properly
 	if sessionIndex == 0 {
-		return &CommonMatchmakeSession{}, fmt.Errorf("No gatherings available!"), nex.Errors.RendezVous.LimitExceeded
+		CurrentGatheringID = nex.NewCounter(0)
+		sessionIndex = GetAvailableGatheringID()
 	}
 
 	session := CommonMatchmakeSession{
@@ -156,15 +154,16 @@ func CreateSessionByMatchmakeSession(matchmakeSession *match_making_types.Matchm
 		GameMatchmakeSession:   matchmakeSession,
 	}
 
-	Sessions[sessionIndex] = &session
-	Sessions[sessionIndex].GameMatchmakeSession.Gathering.ID = sessionIndex
-	Sessions[sessionIndex].GameMatchmakeSession.Gathering.OwnerPID = hostPID
-	Sessions[sessionIndex].GameMatchmakeSession.Gathering.HostPID = hostPID
+	session.GameMatchmakeSession.Gathering.ID = sessionIndex
+	session.GameMatchmakeSession.Gathering.OwnerPID = hostPID
+	session.GameMatchmakeSession.Gathering.HostPID = hostPID
 
-	Sessions[sessionIndex].GameMatchmakeSession.StartedTime = nex.NewDateTime(0)
-	Sessions[sessionIndex].GameMatchmakeSession.StartedTime.UTC()
-	Sessions[sessionIndex].GameMatchmakeSession.SessionKey = make([]byte, 32)
-    rand.Read(Sessions[sessionIndex].GameMatchmakeSession.SessionKey)
+	session.GameMatchmakeSession.StartedTime = nex.NewDateTime(0)
+	session.GameMatchmakeSession.StartedTime.UTC()
+	session.GameMatchmakeSession.SessionKey = make([]byte, 32)
+	rand.Read(session.GameMatchmakeSession.SessionKey)
+
+	Sessions[sessionIndex] = &session
 
 	return Sessions[sessionIndex], nil, 0
 }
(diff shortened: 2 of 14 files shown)


globals/matchmaking_globals.go
```diff
@@ -14,3 +14,4 @@ type CommonMatchmakeSession struct {
 
 var Sessions map[uint32]*CommonMatchmakeSession
 var CurrentGatheringID = nex.NewCounter(0)
+var CurrentMatchmakingCallID uint32
```
DaniElectra: This can be a counter

globals/matchmaking_utils.go
```diff
@@ -140,6 +142,33 @@ func RemoveClientFromAllSessions(client *nex.Client) {
...
+
+	Sessions[sessionIndex] = &session
+	Sessions[sessionIndex].GameMatchmakeSession.Gathering.ID = sessionIndex
+	Sessions[sessionIndex].GameMatchmakeSession.Gathering.OwnerPID = hostPID
+	Sessions[sessionIndex].GameMatchmakeSession.Gathering.HostPID = hostPID
+
+	Sessions[sessionIndex].GameMatchmakeSession.StartedTime = nex.NewDateTime(0)
+	Sessions[sessionIndex].GameMatchmakeSession.StartedTime.UTC()
+	Sessions[sessionIndex].GameMatchmakeSession.SessionKey = make([]byte, 32)
+    rand.Read(Sessions[sessionIndex].GameMatchmakeSession.SessionKey)
```
DaniElectra: Use tabs, not spaces

globals/matchmaking_utils.go
```diff
@@ -140,6 +142,33 @@ func RemoveClientFromAllSessions(client *nex.Client) {
 	}
 }
 
+// CreateSessionByMatchmakeSession creates a gathering from a MatchmakeSession
+func CreateSessionByMatchmakeSession(matchmakeSession *match_making_types.MatchmakeSession, searchMatchmakeSession *match_making_types.MatchmakeSession, hostPID uint32) (*CommonMatchmakeSession, error, uint32) {
+	sessionIndex := GetAvailableGatheringID()
+	// This should in theory be impossible, as there aren't enough PIDs creating sessions to fill the uint32 limit.
+	// If we ever get here, we must be not deleting sessions properly
```
DaniElectra: This comment does not apply anymore, and can be removed

globals/matchmaking_utils.go
```diff
@@ -168,6 +197,34 @@ func FindSessionByMatchmakeSession(searchMatchmakeSession *match_making_types.Ma
...
+
+	Sessions[sessionIndex] = &session
+	Sessions[sessionIndex].GameMatchmakeSession.Gathering.ID = sessionIndex
+	Sessions[sessionIndex].GameMatchmakeSession.Gathering.OwnerPID = hostPID
+	Sessions[sessionIndex].GameMatchmakeSession.Gathering.HostPID = hostPID
+
+	Sessions[sessionIndex].GameMatchmakeSession.StartedTime = nex.NewDateTime(0)
+	Sessions[sessionIndex].GameMatchmakeSession.StartedTime.UTC()
+	Sessions[sessionIndex].GameMatchmakeSession.SessionKey = make([]byte, 32)
+    rand.Read(Sessions[sessionIndex].GameMatchmakeSession.SessionKey)
```
DaniElectra: Use tabs, not spaces

matchmake-extension/auto_matchmake_postpone.go
```diff
@@ -90,57 +81,12 @@ func autoMatchmake_Postpone(err error, client *nex.Client, callID uint32, anyGat
 	responsePacket.SetSource(0xA1)
 	responsePacket.SetDestination(0xAF)
 	responsePacket.SetType(nex.DataPacket)
-	responsePacket.SetPayload(rmcResponseBytes)
+	responsePacket.SetPayload(rmcResponseBytes)
```
DaniElectra: remove the tab at the end lmao

matchmake-extension/auto_matchmake_with_param_postpone.go
```diff
@@ -0,0 +1,131 @@
...
+	var session *common_globals.CommonMatchmakeSession
+
+	if sessionIndex == 0 {
+		var errCode uint32
+		session, err, errCode = common_globals.CreateSessionByMatchmakeSession(matchmakeSession, searchMatchmakeSession, client.PID())
+		if err != nil {
+			logger.Error(err.Error())
+			return errCode
+		}
+	}else{
```
DaniElectra:
```suggestion
	} else {
```

globals/matchmaking_utils.go
```diff
@@ -140,6 +142,33 @@ func RemoveClientFromAllSessions(client *nex.Client) {
...
 }
 
+// CreateSessionByMatchmakeSession creates a gathering from a MatchmakeSession
+func CreateSessionByMatchmakeSession(matchmakeSession *match_making_types.MatchmakeSession, searchMatchmakeSession *match_making_types.MatchmakeSession, hostPID uint32) (*CommonMatchmakeSession, error, uint32) {
+	sessionIndex := GetAvailableGatheringID()
+	// This should in theory be impossible, as there aren't enough PIDs creating sessions to fill the uint32 limit.
+	// If we ever get here, we must be not deleting sessions properly
+	if sessionIndex == 0 {
+		return &CommonMatchmakeSession{}, fmt.Errorf("No gatherings available!"), nex.Errors.RendezVous.LimitExceeded
+	}
```
jonbarrow: Not an "issue" per say, but since we're already here refactoring stuff. I've been thinking about this, maybe we should just reset the counter instead of erroring here? I think it's a pretty safe assumption that if the uint32 limit is hit then the oldest sessions are very much dead (hell maybe even the people who made them are too)

matchmake-extension/auto_matchmake_with_param_postpone.go
```diff
@@ -0,0 +1,131 @@
...
+		session.GameMatchmakeSession.MatchmakeParam.Parameters["@Lat"].Float64 = 0
+	
+		session.GameMatchmakeSession.MatchmakeParam.Parameters["@CC"] = nex.NewVariant()
+		session.GameMatchmakeSession.MatchmakeParam.Parameters["@CC"].TypeID = 4
+		session.GameMatchmakeSession.MatchmakeParam.Parameters["@CC"].Str = "US"
+	
+		session.GameMatchmakeSession.MatchmakeParam.Parameters["@DR"] = nex.NewVariant()
+		session.GameMatchmakeSession.MatchmakeParam.Parameters["@DR"].TypeID = 1
+		session.GameMatchmakeSession.MatchmakeParam.Parameters["@DR"].Int64 = 0
+	}*/
```
DaniElectra: Why is this commented out? Can it be removed?

globals/matchmaking_utils.go
```diff
@@ -140,6 +142,33 @@ func RemoveClientFromAllSessions(client *nex.Client) {
...
+	Sessions[sessionIndex] = &session
+	Sessions[sessionIndex].GameMatchmakeSession.Gathering.ID = sessionIndex
+	Sessions[sessionIndex].GameMatchmakeSession.Gathering.OwnerPID = hostPID
+	Sessions[sessionIndex].GameMatchmakeSession.Gathering.HostPID = hostPID
+
+	Sessions[sessionIndex].GameMatchmakeSession.StartedTime = nex.NewDateTime(0)
+	Sessions[sessionIndex].GameMatchmakeSession.StartedTime.UTC()
+	Sessions[sessionIndex].GameMatchmakeSession.SessionKey = make([]byte, 32)
+    rand.Read(Sessions[sessionIndex].GameMatchmakeSession.SessionKey)
+
```
jonbarrow: Possible micro-optimization here as well. Instead of doing many index lookups and then modifying the data, we could instead just set all the data on the `session` struct right away and only assign it to `Sessions` later. Unsure how much of a difference that would make but logically it should be faster?

globals/matchmaking_utils.go
```diff
@@ -140,6 +142,33 @@ func RemoveClientFromAllSessions(client *nex.Client) {
...
+	return Sessions[sessionIndex], nil, 0
```
jonbarrow, line 168: Now that we have `MutexMap` in `nex-go` I wonder if that would be safer here rather than a normal `map`

globals/matchmaking_utils.go
```diff
@@ -292,6 +349,185 @@ func AddPlayersToSession(session *CommonMatchmakeSession, connectionIDs []uint32
...
+
+		messagePacket.AddFlag(nex.FlagNeedsAck)
+		messagePacket.AddFlag(nex.FlagReliable)
+
+		server.Send(messagePacket)
+	}
+
+	//This appears to be correct. Tri-Force Heroes uses 3.9.0, and has issues if this is ran.
+	//Minecraft, however, requires this to be ran.
+	//TODO: Check other games both pre and post 3.10.0 and validate.
```
DaniElectra:
```suggestion
	// This appears to be correct. Tri-Force Heroes uses 3.9.0, and has issues if these notifications are sent
	// Minecraft, however, requires this to be sent
	// TODO: Check other games both pre and post 3.10.0 and validate
```

globals/matchmaking_utils.go
```diff
@@ -90,7 +91,8 @@ func RemoveClientFromAllSessions(client *nex.Client) {
 
 			rmcMessage := nex.NewRMCRequest()
 			rmcMessage.SetProtocolID(notifications.ProtocolID)
-			rmcMessage.SetCallID(0xffff0000)
+			rmcMessage.SetCallID(CurrentMatchmakingCallID)
+			CurrentMatchmakingCallID+=1
```
jonbarrow: Can just use the `Counter` type from `nex-go` here instead of managing this yourself

globals/matchmaking_utils.go
```diff
@@ -277,7 +334,7 @@ func FindSessionsByMatchmakeSessionSearchCriterias(lstSearchCriteria []*match_ma
 
 // AddPlayersToSession updates the given sessions state to include the provided connection IDs
 // Returns a NEX error code if failed
-func AddPlayersToSession(session *CommonMatchmakeSession, connectionIDs []uint32) (error, uint32) {
+func AddPlayersToSession(session *CommonMatchmakeSession, connectionIDs []uint32, initiatingClient *nex.Client) (error, uint32) {
```
DaniElectra: We also need to pass the join message now that we are handling notifications here

globals/matchmaking_utils.go
```diff
@@ -292,6 +349,185 @@ func AddPlayersToSession(session *CommonMatchmakeSession, connectionIDs []uint32
...
+			oEvent.StrParam = ""
+			oEvent.Param3 = uint32(len(connectionIDs))
+
+			notificationStream := nex.NewStreamOut(server)
+
+			notificationStream.WriteStructure(oEvent)
+
+			notificationRequestMessage.SetParameters(notificationStream.Bytes())
+			notificationRequestBytes := notificationRequestMessage.Bytes()
+			//fmt.Println(hex.EncodeToString(notificationRequestBytes))
```
DaniElectra:
```suggestion
```

DaniElectra (changes requested):

globals/matchmaking_utils.go
```diff
@@ -5,6 +5,7 @@ import (
 	"math"
 	"strconv"
 	"strings"
+	"math/rand"
```
DaniElectra: can we use `crypto/rand`?

globals/matchmaking_utils.go
```diff
@@ -292,6 +347,181 @@ func AddPlayersToSession(session *CommonMatchmakeSession, connectionIDs []uint32
...
+		oEvent.Param3 = uint32(len(connectionIDs))
+
+		notificationStream := nex.NewStreamOut(server)
+
+		notificationStream.WriteStructure(oEvent)
+
+		notificationRequestMessage.SetParameters(notificationStream.Bytes())
+		notificationRequestBytes := notificationRequestMessage.Bytes()
+
+		target := server.FindClientFromPID(uint32(session.GameMatchmakeSession.Gathering.OwnerPID))
```
DaniElectra: We have to check if `target` is nil, like we are doing on the other notifications

matchmake-extension/auto_matchmake_with_param_postpone.go
shutterbug2000: This was for Tri-Force Heroes, I kept it commented in case it was required, but it's not :p

globals/matchmaking_utils.go
shutterbug2000: I tried to do this, but wasn't sure how MutexMap is meant to work. Might look into it in a bit, but wanted to get the other changes pushed first :p

matchmake-extension/join_matchmake_session_with_param.go
```diff
@@ -23,8 +21,7 @@ func joinMatchmakeSessionWithParam(err error, client *nex.Client, callID uint32,
 	}
 
 	// TODO - More checks here
-
-	err, errCode := common_globals.AddPlayersToSession(session, []uint32{client.ConnectionID()})
+	err, errCode := common_globals.AddPlayersToSession(session, []uint32{client.ConnectionID()}, client, "")
```
DaniElectra: The join message exists on `joinMatchmakeSessionParam`

matchmake-extension/create_matchmake_session_with_param.go
```diff
@@ -22,45 +20,21 @@ func createMatchmakeSessionWithParam(err error, client *nex.Client, callID uint3
...
-		SearchCriteria:       make([]*match_making_types.MatchmakeSessionSearchCriteria, 0),
-		GameMatchmakeSession: joinedMatchmakeSession,
+	session, err, errCode := common_globals.CreateSessionBySearchCriteria(joinedMatchmakeSession, make([]*match_making_types.MatchmakeSessionSearchCriteria, 0), client.PID())
+	if err != nil {
+		logger.Error(err.Error())
+		return errCode
 	}
 
-	err, errCode := common_globals.AddPlayersToSession(common_globals.Sessions[sessionIndex], []uint32{client.ConnectionID()})
+	err, errCode = common_globals.AddPlayersToSession(session, []uint32{client.ConnectionID()}, client, "")
```
DaniElectra: The join message exists inside createMatchmakeSessionParam

globals/matchmaking_utils.go
jonbarrow: `MutexMap` is used to create concurrency-safe maps. If we have multiple threads/go routines reading and writing to the map at the same time, Go will panic

`MutexMap` is a generic type which provides a wrapper around a `map` and uses mutex operations to ensure only one thread/go routine accesses the map at a time

Basic usage would be

```go
func main() {
	mutexMap := NewMutexMap[keyType, valueType]()

	mutexMap.Set(key, value)
	value, ok := mutexMap.Get(key)
	mutexMap.Delete(key)
}
```

globals/matchmaking_utils.go
DaniElectra: Also, there's a space at the end

DaniElectra:
Works with Mario Kart 7 :+1:

globals/matchmaking_globals.go
```diff
@@ -14,3 +14,4 @@ type CommonMatchmakeSession struct {
 
 var Sessions map[uint32]*CommonMatchmakeSession
 var CurrentGatheringID = nex.NewCounter(0)
+var CurrentMatchmakingCallID = nex.NewCounter(0)
```
DaniElectra, line 17: The `CurrentMatchmakingCallID` should also be used on `UnregisterGathering`, `EndParticipation` and other places that use notifications

matchmake-extension/auto_matchmake_with_param_postpone.go
```diff
@@ -0,0 +1,85 @@
...
+	// A client may disconnect from a session without leaving reliably,
+	// so let's make sure the client is removed from the session
+	common_globals.RemoveClientFromAllSessions(client)
+
+	var matchmakeSession *match_making_types.MatchmakeSession
+	matchmakeSession = autoMatchmakeParam.SourceMatchmakeSession
+
+	searchMatchmakeSession := matchmakeSession.Copy().(*match_making_types.MatchmakeSession)
+	commonMatchmakeExtensionProtocol.cleanupSearchMatchmakeSessionHandler(searchMatchmakeSession)
+	sessionIndex := common_globals.FindSessionByMatchmakeSession(searchMatchmakeSession)
```
DaniElectra: `AutoMatchmakeWithParam_Postpone` has search criteria, so I would find the session using `FindSessionsByMatchmakeSessionSearchCriterias`

jonbarrow:
Confirmed working in Mario Kart 8 :+1:

shutterbug2000:
This has now been testing working in Splatoon

globals/matchmaking_utils.go
DaniElectra: The real map isn't exposed but we have to range through it, so we can't use that atm. Using `Each` is not an option, since on `FindClientSession` we may return early while iterating

DaniElectra:
@jonbarrow Any objections with merging this?

jonbarrow:
Nope! Looks good to me :+1: I see it's still a draft, so once it's marked as ready it can be merged

shutterbug2000:
> Nope! Looks good to me 👍 I see it's still a draft, so once it's marked as ready it can be merged

I've marked it as ready!

DaniElectra (approved):

PR #22, opened 2023-10-19
Redesign find sessions using SearchCriteria

DaniElectra:
The `MatchmakeSessionSearchCriteria` is intended to be compared with the original session to find valid sessions. Thus, we can remove the `SearchCriteria` field that we are storing internally and instead compare values with the session.

With this, we implement comparing attributes by default, unless game-specific checks are defined, in which case the attribute search gets overriden. This allows overriding region-locked matchmaking.

This also allows removing session creation that requires search criteria `CreateSessionBySearchCriteria` and instead move everything to `CreateSessionByMatchmakeSession`.

We can also remove the cleanup of search criteria, since it's redundant considering we allow game-specific checks.

Friends-only matches are also now supported, and requires setting the `GetUserFriendPIDs` handler on `MatchmakeExtension`.

This improves accuracy finding sessions and is how Rambo's MK8 server is implemented.

Aside from that, I'm considering moving the logger to the globals, so that we can log from `globals`.

Tested successfully with:
- Mario Kart 7
- Steel Diver: Sub Wars
- IRONFALL Invasion

### DIFF
--- a/globals/matchmaking_utils.go
+++ b/globals/matchmaking_utils.go
@@ -111,8 +111,7 @@ func RemoveClientFromAllSessions(client *nex.Client) {
 
 			targetClient := server.FindClientFromPID(uint32(ownerPID))
 			if targetClient == nil {
-				// TODO - We don't have a logger here
-				// logger.Warning("Owner client not found")
+				Logger.Warning("Owner client not found")
 				gid = FindClientSession(client.ConnectionID())
 				continue
 			}
@@ -211,8 +210,7 @@ func FindSessionByMatchmakeSession(pid uint32, searchMatchmakeSession *match_mak
 		// TODO - Is this a flag or a constant?
 		if sessionToCheck.GameMatchmakeSession.ParticipationPolicy == 98 {
 			if GetUserFriendPIDsHandler == nil {
-				// TODO - We don't have a logger here
-				// logger.Warning("Missing GetUserFriendPIDsHandler!")
+				Logger.Warning("Missing GetUserFriendPIDsHandler!")
 				continue
 			}
 
@@ -272,8 +270,7 @@ func FindSessionsByMatchmakeSessionSearchCriterias(pid uint32, lstSearchCriteria
 			// TODO - Is this a flag or a constant?
 			if session.GameMatchmakeSession.ParticipationPolicy == 98 {
 				if GetUserFriendPIDsHandler == nil {
-					// TODO - We don't have a logger here
-					// logger.Warning("Missing GetUserFriendPIDsHandler!")
+					Logger.Warning("Missing GetUserFriendPIDsHandler!")
 					continue
 				}
 
@@ -364,7 +361,7 @@ func AddPlayersToSession(session *CommonMatchmakeSession, connectionIDs []uint32
 		target := server.FindClientFromConnectionID(session.ConnectionIDs[i])
 		if target == nil {
 			// TODO - Error here?
-			//logger.Warning("Player not found")
+			Logger.Warning("Player not found")
 			continue
 		}
 
@@ -420,7 +417,7 @@ func AddPlayersToSession(session *CommonMatchmakeSession, connectionIDs []uint32
 			target := server.FindClientFromConnectionID(session.ConnectionIDs[i])
 			if target == nil {
 				// TODO - Error here?
-				//logger.Warning("Player not found")
+				Logger.Warning("Player not found")
 				continue
 			}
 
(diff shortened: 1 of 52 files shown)


jonbarrow:
I will give this a deeper review later on Friday when I'm back home. However I will say now, I agree with moving the logger to the globals. Not having access to the logger in some cases is not great

globals/matchmaking_utils.go
```diff
@@ -163,13 +163,29 @@ func CreateSessionByMatchmakeSession(matchmakeSession *match_making_types.Matchm
...
+		session.GameMatchmakeSession.MatchmakeParam.Parameters = make(map[string]*nex.Variant)
+	}
+
+	session.GameMatchmakeSession.MatchmakeParam.Parameters["@SR"] = nex.NewVariant()
+	session.GameMatchmakeSession.MatchmakeParam.Parameters["@SR"].TypeID = 3
+	session.GameMatchmakeSession.MatchmakeParam.Parameters["@SR"].Bool = true
+
+	session.GameMatchmakeSession.MatchmakeParam.Parameters["@GIR"] = nex.NewVariant()
+	session.GameMatchmakeSession.MatchmakeParam.Parameters["@GIR"].TypeID = 1
+	session.GameMatchmakeSession.MatchmakeParam.Parameters["@GIR"].Int64 = 3
```
shutterbug2000, line 179: This is more just a curiosity than anything, but do we know what purpose these serve?
The official NEX server assigns them (and the Tri-Force Heroes server assigns even more than this), but I've yet to find any client using (or requiring) them.

DaniElectra: I don't think we know the exact purpose, but they are assigned by the server when creating a session on every title I remember so far

DaniElectra:
> I agree with moving the logger to the globals. Not having access to the logger in some cases is not great

Alright, the loggers have been moved to the globals!

globals/matchmaking_utils.go
jonbarrow: I know that Modern Warfare II uses `SR` to mean "Skill Rating" which is used to determine which rank you are and get matched with. I wonder if this is something similar? I know MW2 is not a NEX, or even WiiU, game but I wonder if the terminology is the same/similar? `@SR` being `true` meaning that "Skill Rating" is used when match making, and `@GIR` (**G**athering IDK **R**ating?) is whatever rank?

globals/matchmaking_utils.go
```diff
@@ -190,144 +206,135 @@ func FindSessionByMatchmakeSession(searchMatchmakeSession *match_making_types.Ma
 			continue
 		}
 
+		// If the session only allows friends, check if the owner is in the friend list of the PID
+		// TODO - Is this a flag or a constant?
+		if sessionToCheck.GameMatchmakeSession.ParticipationPolicy == 98 {
+			if GetUserFriendPIDsHandler == nil {
+				Logger.Warning("Missing GetUserFriendPIDsHandler!")
+				continue
```
jonbarrow, line 214: Should we maybe bail here, like the other handler checks, instead of continuing?

jonbarrow (commented):
Looks really solid from what I can tell!

globals/matchmaking_utils.go
DaniElectra: Since we are ranging every session, continuing allows public matches to work even if friend matches are not implemented.

This could be useful in the weird edge case where we discover a game uses friend matches after we implement the server. If we didn't do this, any public matches created after the friend match wouldn't show up.

globals/matchmaking_utils.go
DaniElectra: > I know that Modern Warfare II uses `SR` to mean "Skill Rating" which is used to determine which rank you are and get matched with. I wonder if this is something similar? I know MW2 is not a NEX, or even WiiU, game but I wonder if the terminology is the same/similar? `@SR` being `true` meaning that "Skill Rating" is used when match making, and `@GIR` (**G**athering IDK **R**ating?) is whatever rank?

I could buy that if the types were the same, but that isn't the case. It could be something similar though

globals/matchmaking_utils.go
jonbarrow: :+1:

shutterbug2000:
No issues that I can see! (LGTM ;P)
Is it ready to merge? Have some (unrelated) bugfixes I want to push, but don't want to cause any conflicts :p

DaniElectra:
> Is it ready to merge?

Yes, it should be! I'll wait if @jonbarrow has anything else to say and then it can be merged

jonbarrow:
Merge away 👍

PR #24, opened 2023-10-29
Add common DataStore protocol

jonbarrow:
Lays the ground work for implementing a common DataStore protocol. This one is going to be very heavy on helper methods just due to the nature of how DataStore needs to operate

So far this implements enough for Super Mario Maker, every stock DataStore method I've seen the game use is implemented here. All methods *seen in Super Mario Maker* have been heavily tested using NintendoClients to verify the functionality against the real server including many edge cases, and I have tried to leave detailed comments in areas that may have confusing handling

Note that this is far from a complete implementation at this time. Some additional methods outside of the ones seen in Super Mario Maker are also implemented, however they are ported from existing servers which use DataStore. These servers are older and the functionality of these methods has not been 100% tested, but it all *seems* correct

One thing we should also decide on now is how to handle S3. For several methods we need access to the servers S3 bucket. This is currently just set using `SetS3Bucket`. However we also have the option of just having the user pass some configuration parameters and then the *common* library creates and manages the S3 client. Originally this would not be done in order to give the developer as much control as possible. However this *will* lead to some, honestly unnecessary, duplicated code between server implementations. The same initialization and helper methods will be defined in basically all servers. And with S3 options now extremely limited, I'm not positive this is a worthwhile thing to support. I'm not a fan of vendor lock-in, but when it comes to DataStore there isn't much of a choice now besides MinIO. Trying to use the major cloud providers is no longer tenable, and I plan to recommend the use of MinIO in all servers which use DataStore in the repo's `README` anyway, plus the Super Mario Maker server already just uses the MinIO SDK

Moving the S3 client to common (and providing a getter for it for game-specific stuff) would remove the need to set several helper methods as well, however I understand if there is opposition to this idea as it *DOES* enter the realm of vendor lock-in

I do not think we should control the database layer though, the proposal is only for the S3 layer. We make use of Postgres to keep things uniform across our servers, and that code will be duplicated, but there is a legitimate use case for not controlling that layer here. Other developers are not limited in terms of database solutions as they are with S3 options, so anyone wanting to fork and rehost our servers should always be allowed to swap out Postgres with whatever database solution they use so long as the helper functions are implemented correctly

Currently implements

- `DataStore::DeleteObject`. Ported from MVDK:TS server. Functionality not 100% verified
- `DataStore::GetMeta`. Tested against SMM. Functionality (mostly) verified
- `DataStore::GetMetas`. Ported from MVDK:TS server. Functionality not 100% verified
- `DataStore::SearchObject`. Ported from SM3DW server. Functionality not 100% verified
- `DataStore::RateObject`. Ported from MVDK:TS server. Functionality not 100% verified
- `DataStore::PostMetaBinary`. Ported from MVDK:TS server. Functionality not 100% verified
- `DataStore::PreparePostObject`. Tested against SMM. Functionality (mostly) verified
- `DataStore::PrepareGetObject`. Tested against SMM. Functionality (mostly) verified
- `DataStore::CompletePostObject`. Tested against SMM. Functionality (mostly) verified
- `DataStore::GetMetasMultipleParam`. Tested against SMM. Functionality (mostly) verified
- `DataStore::CompletePostObjects`. Ported from MVDK:TS server. Functionality not 100% verified
- `DataStore::ChangeMeta`. Tested against SMM. Functionality (mostly) verified
- `DataStore::RateObjects`. Tested against SMM. Functionality (mostly) verified


datastore/complete_post_object.go
```diff
@@ -0,0 +1,86 @@
...
+		common_globals.Logger.Warning("UpdateObjectUploadCompletedByDataID not defined")
+		return nex.Errors.Core.NotImplemented
+	}
+
+	if err != nil {
+		common_globals.Logger.Error(err.Error())
+		return nex.Errors.DataStore.Unknown
+	}
+
+	// TODO - What is param.IsSuccess? Is this correct?
```
DaniElectra: The param `isSuccess` determines if the client could not make the S3 request or the S3 server returned an error

datastore/complete_post_object.go
```diff
@@ -0,0 +1,86 @@
...
+	if param.IsSuccess {
+		bucket := commonDataStoreProtocol.s3Bucket
+		key := fmt.Sprintf("%d.bin", param.DataID)
```
DaniElectra: I don't think we want to store all files on a root directory, since we want to support HPP. I think it would be best to split this data on two folders (`data` and `notify`, for example), the last one would be used by HPP for giving notification URLs to the client.

datastore/complete_post_object.go
```diff
@@ -0,0 +1,86 @@
...
+
+		objectSizeS3, err := commonDataStoreProtocol.s3ObjectSizeHandler(bucket, key)
+		if err != nil {
+			common_globals.Logger.Error(err.Error())
+			return nex.Errors.DataStore.NotFound
+		}
+
+		objectSizeDB, errCode := commonDataStoreProtocol.getObjectSizeDataIDHandler(param.DataID)
```
DaniElectra: do we want to store the object size locally?

datastore/prepare_get_object.go
```diff
@@ -0,0 +1,98 @@
...
+	if errCode != 0 {
+		return errCode
+	}
+
+	errCode = commonDataStoreProtocol.verifyObjectPermissionHandler(objectInfo.OwnerID, client.PID(), objectInfo.Permission)
+	if errCode != 0 {
+		return errCode
+	}
+
+	URL, err := commonDataStoreProtocol.presignGetObjectHandler(bucket, key, time.Minute*15)
```
DaniElectra: This sould be lowercase `url`

jonbarrow:
Marking this as ready for review now because all DataStore methods that I have personally seen in servers has been implemented

Since this PR comes from a branch on this repo though please feel free to push any new methods for games I have not touched

datastore/search_object.go
```diff
@@ -0,0 +1,106 @@
...
+	objects, totalCount, errCode := commonDataStoreProtocol.getObjectInfosByDataStoreSearchParamHandler(param)
+	if errCode != 0 {
+		return errCode
+	}
+
+	pSearchResult := datastore_types.NewDataStoreSearchResult()
+
+	pSearchResult.TotalCount = totalCount
+	pSearchResult.Result = make([]*datastore_types.DataStoreMetaInfo, 0, len(objects))
+	pSearchResult.TotalCountType = uint8(param.DataType) // TODO - Is this right?
```
DaniElectra: No, this represents if all results are returned or not (see https://nintendo-wiki.pretendo.network/docs/nex/protocols/datastore/#datastoresearchresult-structure). It's worth noting that the values on the wiki *could* be flags (0x1: Not all results, 0x2: Disabled), haven't checked though

datastore/get_metas.go
```diff
@@ -0,0 +1,159 @@
...
+	// * param has an AccessPassword, but it goes unchecked here.
+	// * The password would need to be the same for every object
+	// * in the input array, which doesn't make any sense. Assuming
+	// * it's unused until proven otherwise
+
+	for i := 0; i < len(dataIDs); i++ {
+		objectInfo, errCode := commonDataStoreProtocol.getObjectInfoByDataIDHandler(dataIDs[i])
+
+		if errCode != 0 {
+			// TODO - Maybe this should be broken out into a util function in globals?
```
DaniElectra: I agree, I think we could make a helper that creates an empty `DataStoreMetaInfo`

DaniElectra (commented):
I'm personally fine with moving the S3 client to the common protocols. Using the MinIO SDK shouldn't affect implementations that want to use another server since they all use S3. In order to allow custom configurations, I think we could let the server implementation create the MinIO client and pass it to the common protocols.

Also, we should consider integrating HPP here aswell, since the implementation seems to be fairly standard across all my dumps. For reference, these are the methods that it uses:
- [PrepareGetObjectV1](https://nintendo-wiki.pretendo.network/docs/nex/protocols/datastore/#1-preparegetobjectv1)
- [PreparePostObjectV1](https://nintendo-wiki.pretendo.network/docs/nex/protocols/datastore/#2-preparepostobjectv1)
- [CompletePostObjectV1](https://nintendo-wiki.pretendo.network/docs/nex/protocols/datastore/#3-completepostobjectv1)
- [GetNotificationUrl](https://nintendo-wiki.pretendo.network/docs/nex/protocols/datastore/#13-getnotificationurl)
- [GetNewArrivedNotificationsV1](https://nintendo-wiki.pretendo.network/docs/nex/protocols/datastore/#14-getnewarrivednotificationsv1)
- [GetSpecificMetaV1](https://nintendo-wiki.pretendo.network/docs/nex/protocols/datastore/#20-getspecificmetav1)

This doesn't affect the currently implemented methods on this PR, but it's worth mentioning since games like MK7 will use V1 versions of methods. On all HPP requests I've seen, the flags of `PreparePostParamV1` is always 0x8, so I think we could check for that when handling notifications (unless we find a game that uses the flag outside of HPP).

datastore/complete_post_object.go
jonbarrow: Are there any games which actually behave this way? Super Mario Maker simply throws up an error when the S3 upload fails. I don't recall what Tipping Stars does in these cases, and SM3DW doesn't post objects. Those are the only games I've worked on that use DataStore

If that's how the client is supposed to behave though then I believe the implementation is correct no? The comment just needs to be removed?

datastore/complete_post_object.go
jonbarrow: I'm not versed in how HPP operates, why does it need it's own folder? I'm not opposed to this change (SMM originally used the folder `10.WUP_AMAJ_datastore/ds/1/data`, so it's not unheard of), I just wanna make sure I understand the usecase

datastore/complete_post_object.go
jonbarrow: It provides a much quicker way to get the objects size, which is needed when sending download URLs to the client, since it can be retrieved with a single SQL query rather than querying S3 every time. Also it allows us to verify that the object posted correctly/completely, just in case some weird edge case happens where the upload only partially completes

datastore/complete_post_object.go
DaniElectra: In Badge Arcade if the upload fails, it will send the `CompletePostObject` without the `isSuccess` param set, and then show an error to the user.

> If that's how the client is supposed to behave though then I believe the implementation is correct no? The comment just needs to be removed?

Yes, this should be correct, although I think we should remove the data ID from the server if the operation fails (this is after checking that the data ID actually doesn't exist on S3)

datastore/search_object.go
jonbarrow: Oh perfect. So I assume that if `len(objects)` matches `totalCount` then `TotalCountType` would be 0, and if not then it would be 1? What is the purpose of option 3 if you know? I'm unsure of when the client would expect this to be disabled? Or how to even check for that. And do we know if an option 2 exists?

datastore/get_metas.go
jonbarrow: I was thinking about this just now, it would take a lot of work to go through all the types but this should honestly probably be implemented in `nex-protocols-go`. Right now all the types `New` functions return zeroed structs, which works fine for primative types like numbers and strings. Those will default to their zero. But in structs which have fields that are pointers to other structs that's when this breaks down since that results in a nil pointer

The `New` functions should honestly be updated to just return properly zeroed structs rather than ones containing nil pointers. That way we can just call the `New` function to get our empty structs

datastore/search_object.go
DaniElectra: > Oh perfect. So I assume that if `len(objects)` matches `totalCount` then `TotalCountType` would be 0, and if not then it would be 1?

Yes, that should be it.

> What is the purpose of option 3 if you know?

The option 3 can be trigegred if the `totalCountEnabled` option in the search params ise set to false. Note that this is only available on revision 3 of the structure or NEX 4.0+

> And do we know if an option 2 exists?

I'm not sure since I haven't seen it

jonbarrow:
> Using the MinIO SDK shouldn't affect implementations that want to use another server since they all use S3

Oh perfect. The MinIO docs have separate examples for when connecting to MinIO vs AWS so I assumed the MinIO client could only connect to MinIO servers. Awesome

> In order to allow custom configurations, I think we could let the server implementation create the MinIO client and pass it to the common protocols

This works for me. I was originally thinking about just providing a custom struct the user could fill their config details into and then common would create the client. But that was somewhat naive, and ignored advanced configurations. This is the better solution

> we should consider integrating HPP here aswell

I agree. As far as I know HPP should be, for servers which support it, a 1:1 substitute for NEX, and any protocols provided by NEX would be available to the HPP client (correct me if I'm wrong there). So it being supported here makes sense

What all work needs to be done for this? I haven't looked much into HPP, even the implementation in `nex-go`, so I'm unsure what changes need to be made to bring support here

> MK7 will use V1 versions of methods

Over HPP or over NEX? Either way those should be implemented here tbh, we can add these to this PR if we have working implemenations of them. Just push to the branch

> On all HPP requests I've seen, the flags of `PreparePostParamV1` is always 0x8, so I think we could check for that when handling notifications (unless we find a game that uses the flag outside of HPP)

I'm not positive that's what this flag means, but it sounds like a good assumption for now. SMM doesn't use this flag for it's objects. Objects in SMM only have flags of 0, 256, or 3840 from what I can tell and I have no idea what they mean. We should make a note in the implementation though that this is basically just a heuristics guess and the exact flag use is not known

datastore/complete_post_object.go
jonbarrow: Love that SMM just doesn't do that lol. Okay sounds good

datastore/complete_post_object.go
DaniElectra: HPP has two subdirectories for handling S3 files: `data` and `notify` (with the main directory being the same `10.WUP_KCFJ_datastore/ds/1`).

The data folder has the meta binary that the client uploads to S3. Then, there is a `notify` folder which the client uses to determine if it has received a new "notification", which would be if someone uploads a meta binary and the client has permission to access it.

The client accesses an S3 notification URL identified with its PID such as `notify/{%011d}`, without the brackets `{}`. The URL is provided by the HPP server. The contents of a notification can be seen [here](https://github.com/PretendoNetwork/pokemon-rumble-world-secure/blob/e6e7aa8a239bbc9743c8866b244344f95ffb3259/nex/datastore/complete_post_object_v1.go#L51).

We would split the data and notifications to avoid potential collisions.

datastore/search_object.go
jonbarrow: I totally missed that field, sweet

datastore/complete_post_object.go
jonbarrow: > HPP has two subdirectories for handling S3 files: `data` and `notify` (with the main directory being the same `10.WUP_KCFJ_datastore/ds/1`).

Okay, I think I'll actually just mimic this structure entirely tbh. I'm going to assume that 3DS titles will use `CTR` instead of `WUP`



> The data folder has the meta binary that the client uploads to S3

I'm assuming this is just what this specific game decides to do, yes? Because in SMM and Tipping Stars this is not the case, the data uploaded to S3 does not match the meta binary. Or does HPP just work completely differently here?

> Then, there is a `notify` folder which the client uses to determine if it has received a new "notification", which would be if someone uploads a meta binary and the client has permission to access it.

Gotcha. This is in stark contrast to how SMM handles notifications since it uses BufferQueue's and object ratings for that (and doesn't use HPP)

> The client accesses an S3 notification URL identified with its PID such as `notify/{%011d}`, without the brackets `{}`. The URL is provided by the HPP server. The contents of a notification can be seen [here](https://github.com/PretendoNetwork/pokemon-rumble-world-secure/blob/e6e7aa8a239bbc9743c8866b244344f95ffb3259/nex/datastore/complete_post_object_v1.go#L51).
> 
> We would split the data and notifications to avoid potential collisions.

So the `notify` content will *ALWAYS* be the users PID? Is this hard coded into the library or does the server provide the *upload* URL (I see you mentioned the accessing URL, but didn't mention uploading)? I agree with splitting, I just want to make sure that the implementation is correct. Right now whenever the client wants to upload an object or a meta binary it creates a new DataStore object for the upload

Are these `notify` files objects as well? Do they get DataIDs like regular objects? Or do we now have to maintain separate tables for standard objects vs notifications?

Also a note on the notification contents you linked

> // TODO - What is the last number? Looks constant per-game

`1425562994` is the unix timestamp for March 5, 2015, which is VERY close to the release date of Rumble World (April 8, 2015). Could very likely be related?

datastore/complete_post_object.go
DaniElectra: I see. In that case, I think we should log an error when the S3 size and the local size doesn't match

DaniElectra:
> What all work needs to be done for this? I haven't looked much into HPP, even the implementation in `nex-go`, so I'm unsure what changes need to be made to bring support here

I was thinking about this and I think we can't differentiate a NEX packet from an HPP packet right now? On `nex-protocols-go` we are checking the packet type, but we don't have access to that here (and there is no param on the server right now that differentiates one from another). I think we could add a boolean on nex-go to check this?

Aside from this, I was thinking that we don't have access to the packet source and destination when we reach a method. We will need to look into this when we want to implement RV.

> Over HPP or over NEX? Either way those should be implemented here tbh, we can add these to this PR if we have working implemenations of them. Just push to the branch

This is over NEX on PRUDP, Mario Kart 7 uses methods like `PreparePostObjectV1` working the same as the normal method.

> I'm not positive that's what this flag means, but it sounds like a good assumption for now. SMM doesn't use this flag for it's objects. Objects in SMM only have flags of 0, 256, or 3840 from what I can tell and I have no idea what they mean. We should make a note in the implementation though that this is basically just a heuristics guess and the exact flag use is not known

Sure, we can add a note about this

datastore/complete_post_object.go
DaniElectra: > Okay, I think I'll actually just mimic this structure entirely tbh. I'm going to assume that 3DS titles will use `CTR` instead of `WUP`

Yes, it's `CTR`. I had copied your SMM example and forgot to rename that 😅

> I'm assuming this is just what this specific game decides to do, yes? Because in SMM and Tipping Stars this is not the case, the data uploaded to S3 does not match the meta binary. Or does HPP just work completely differently here?

Sorry, I was mixing things up. Yes, this is the data that gets uploaded, the meta binary is empty aswell.


> So the `notify` content will _ALWAYS_ be the users PID? Is this hard coded into the library or does the server provide the _upload_ URL (I see you mentioned the accessing URL, but didn't mention uploading)? I agree with splitting, I just want to make sure that the implementation is correct. Right now whenever the client wants to upload an object or a meta binary it creates a new DataStore object for the upload

The S3 key for notifications is the user's PID, just like the S3 key for normal data is the Data ID. The notification is uploaded by the HPP server itself.

> Are these `notify` files objects as well? Do they get DataIDs like regular objects? Or do we now have to maintain separate tables for standard objects vs notifications?
> 
> Also a note on the notification contents you linked
> 
> > // TODO - What is the last number? Looks constant per-game
> 
> `1425562994` is the unix timestamp for March 5, 2015, which is VERY close to the release date of Rumble World (April 8, 2015). Could very likely be related?

This actually could make sense. I checked the Swapdoodle value and it gives the timestamp on November 14 2016, a few days before the first release on (November 17, 2016)

jonbarrow:
> Aside from this, I was thinking that we don't have access to the packet source and destination when we reach a method. We will need to look into this when we want to implement RV.

I was also thinking about this the other day. Easiest solution would probably be to just send the original packet to the method instead of the client. The client can be obtained from the packets `Sender()` method instead, and then we have full access to all the context we've been missing

> This is over NEX on PRUDP, Mario Kart 7 uses methods like `PreparePostObjectV1` working the same as the normal method.

Gotcha gotcha

datastore/complete_post_object.go
jonbarrow: > The notification is uploaded by the HPP server itself

Gotcha gotcha, so we don't have to worry about creating objects in the database for this then. For all intents and purposes it sounds very "internal" so I think we can just keep this separate from the normal objects database. Good to know

datastore/get_metas.go
DaniElectra: I agree, that would be the best solution

datastore/complete_post_object.go
jonbarrow: This should be resolved now. The server now takes in a base for the key used for both data and notify contents

datastore/get_metas.go
jonbarrow: I added this to all the `New` functions in stock DataStore and SMM (since that's what this is aimed at right now). The others can come later as we need them, it's not relevant to add zeroed struct support to every protocol right now

jonbarrow:
@DaniElectra @shutterbug2000 How are we feeling about the proposal to stop sending the client in the protocol handlers and instead sending the original packet? We can still get the client using the packet's `Sender()` method, and it would allow us to:

- Seamlessly support HPP from what I can tell, with no further changes to either the protocols or common libs
- Stop hard-coding the destination and source addresses in preparation for QRV support

I know it would make the function signatures look a bit goofy, but I'm thinking it's a worthwhile change considering the benefits

DaniElectra:
> How are we feeling about the proposal to stop sending the client in the protocol handlers and instead sending the original packet? We can still get the client using the packet's `Sender()` method, and it would allow us to:
> 
>     * Seamlessly support HPP from what I can tell, with no further changes to either the protocols or common libs
> 
>     * Stop hard-coding the destination and source addresses in preparation for QRV support
> 
> 
> I know it would make the function signatures look a bit goofy, but I'm thinking it's a worthwhile change considering the benefits

I agree. Go with it!

jonbarrow:
> > How are we feeling about the proposal to stop sending the client in the protocol handlers and instead sending the original packet? We can still get the client using the packet's `Sender()` method, and it would allow us to:
> 
> > 
> 
> >     * Seamlessly support HPP from what I can tell, with no further changes to either the protocols or common libs
> 
> > 
> 
> >     * Stop hard-coding the destination and source addresses in preparation for QRV support
> 
> > 
> 
> > 
> 
> > I know it would make the function signatures look a bit goofy, but I'm thinking it's a worthwhile change considering the benefits
> 
> 
> 
> I agree. Go with it!

Awesome, I'll get the changes made then! I don't think it's entirely relevant to this PR however so I'll make another one after this PR (seeing as all the protocols here will need updating)

For now, is there anything else left before we merge this?

datastore/protocol.go
```diff
@@ -0,0 +1,281 @@
...
+		return 0
+	}
+
+	// * Allow friends
+	if permission.Permission == 1 {
+		friendsList := c.getUserFriendPIDsHandler(ownerPID)
+
+		if !slices.Contains(friendsList, accessorPID) {
+			return nex.Errors.DataStore.PermissionDenied
+		}
```
DaniElectra, line 83: The friend list will not include the PID that created the object. A separate check should be added

datastore/protocol.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+	// * Allow people in permission.RecipientIDs
+	if permission.Permission == 2 {
+		if !slices.Contains(permission.RecipientIDs, accessorPID) {
+			return nex.Errors.DataStore.PermissionDenied
+		}
```
DaniElectra, line 90: For what I've seen, the recipient ID list doesn't include the PID that creates the object. A separate check should be added

datastore/complete_post_object.go
```diff
@@ -0,0 +1,96 @@
...
+
+		errCode = commonDataStoreProtocol.updateObjectUploadCompletedByDataIDHandler(param.DataID, true)
+		if errCode != 0 {
+			return errCode
+		}
+	} else {
+		errCode := commonDataStoreProtocol.deleteObjectByDataIDHandler(param.DataID)
+		if errCode != 0 {
+			return errCode
+		}
```
DaniElectra, line 95: We have to check if the data ID exists on the S3 server before deleting, just in case

datastore/protocol.go
```diff
@@ -0,0 +1,281 @@
...
+	server                  *nex.Server
+	DefaultProtocol         *datastore.Protocol
+	SuperMarioMakerProtocol *datastore_super_mario_maker.Protocol
+
+	s3Bucket                                            string
+	s3DataKeyBase                                       string
+	s3NotifyKeyBase                                     string
+	rootCACert                                          []byte
+	MinIOClient                                         *minio.Client
+	S3Presigner                                         S3PresignerInterface
```
DaniElectra: do we have to export these?

jonbarrow: These are exported for if a developer wants to use them without having to define them manually in each server. For example in the SMM server there's multiple methods which need presigned URLs so it can just use the exported `S3Presigner` here and not make one itself. Though yeah the `MinIOClient` is less useful exported since the server needs to make it already

datastore/complete_post_object.go
jonbarrow: My intention with `deleteObjectByDataIDHandler` is to remove it from the database, not touching S3 since presumably `IsSuccess` would only be false if the upload failed right? So the file wouldn't be in S3 anyway?

I can still add an additional check for that though if you think it's an issue, or we can delegate that to the server implementing common (not deleting here, but making that part of the implementation of `deleteObjectByDataIDHandler`)

DaniElectra: Is this handler used anywhere else? Since this is an special case, the server implementation should know that the deletion is because of failed upload

jonbarrow: Right now this is only used here in common but there are other places where it will eventually be used. Such as in SMM, which adds methods `PrepareAttachFile` and `CompleteAttachFile`, so this will be reused there. Also there's `DeleteObject` and `DeleteObjects`, though those come with a password so they'll likely get their own deletion functions to handle that. Also the function could be reused in the future in the gRPC API I'm going to eventually add

DaniElectra: Well actually, after thinking more on it, we could just delete the data that is uploaded to S3 aswell. My concern was that the data is uploaded to S3 but the client (or custom client) reported failed upload

We should also check for permissions and if the database doesn't report the upload as complete, but I think that can be done on the delete handler (at least the permissions part)

jonbarrow: > Well actually, after thinking more on it, we could just delete the data that is uploaded to S3 aswell. My concern was that the data is uploaded to S3 but the client (or custom client) reported failed upload

I assumed this to be what you meant originally. So I was clarifying that the intention of `DeleteObjectByDataID` is to remove the object from the database, but not S3 in this function specifically

Most of these helper methods are just me replacing database calls from the SMM implementation with a helper method instead, so that's typically the context they work under unless otherwise specified (such as `S3ObjectSize` which pulls the objects size from S3)

So the question becomes, do we add an additional check and helper method named something like `S3RemoveObject` and call that as well, or do we just keep `DeleteObjectByDataID` and assume that the server implementing that function will do it's own S3 clean up

> We should also check for permissions and if the database doesn't report the upload as complete, but I think that can be done on the delete handler (at least the permissions part)

Unless I'm misunderstanding the suggestion, this is an implementation detail really and is how things are already implemented by SMM. For example in the SMM server the `GetObjectInfoByDataID` is implemented like so

```go
func GetObjectInfoByDataID(dataID uint64) (*datastore_types.DataStoreMetaInfo, uint32) {
	errCode := IsObjectAvailable(dataID)
	if errCode != 0 {
		return nil, errCode
	}

	// The database query handling
}
```

Where `IsObjectAvailable` is implemented as

```go
func IsObjectAvailable(dataID uint64) uint32 {
	var underReview bool

	err := database.Postgres.QueryRow(`SELECT under_review FROM datastore.objects WHERE data_id=$1 AND upload_completed=TRUE AND deleted=FALSE`, dataID).Scan(&underReview)
	if err != nil {
		if err == sql.ErrNoRows {
			return nex.Errors.DataStore.NotFound
		}

		globals.Logger.Error(err.Error())
		// TODO - Send more specific errors?
		return nex.Errors.DataStore.Unknown
	}

	if underReview {
		return nex.Errors.DataStore.UnderReviewing
	}

	return 0
}
```

So whenever a method needs to interact with an object it always verifies the object is available first. This isn't usable here though since `IsObjectAvailable` checks to see if `upload_completed` is true. This is actually the reason why `GetObjectSizeByDataID` exists, since that method specifically does not call `IsObjectAvailable` and always returns the objects size. `GetObjectInfoByDataID` can't be used in this context to get the objects size because the `upload_completed` check will always fail, due to the upload not being marked as completed yet

DaniElectra: > So the question becomes, do we add an additional check and helper method named something like `S3RemoveObject` and call that as well, or do we just keep `DeleteObjectByDataID` and assume that the server implementing that function will do it's own S3 clean up

We can let the implementation handle S3, considering the current implementation of `DeleteObject` lets the handler do that

> > We should also check for permissions and if the database doesn't report the upload as complete, but I think that can be done on the delete handler (at least the permissions part)
> 
> Unless I'm misunderstanding the suggestion, this is an implementation detail really and is how things are already implemented by SMM. 

I mean more like in the deletion segment of `CompletePostObject`, since we only call delete if the upload failed, and nothing else. If there aren't any safety checks on the delete handler, I could send a `CompletePostObject` with any data ID and report as upload failed, and the server would delete the data.

jonbarrow: > I mean more like in the deletion segment of `CompletePostObject`, since we only call delete if the upload failed, and nothing else. If there aren't any safety checks on the delete handler, I could send a `CompletePostObject` with any data ID and report as upload failed, and the server would delete the data.

In the version I have that's not pushed right now I added a helper method to get an objects owner and compare it to the PID of the user making the request, so only an owner of an object can modify it here. But I do think it's a good suggestion to also check if it's been marked as uploaded or not, I'll add that

DaniElectra (approved):
We may want to change the S3 data key to hold the update number in the future, but I think this is fine for now

PR #28, opened 2024-04-10
Refactor of matchmaking_utils.go

PabloMK7:
This change implements the following things:
- Add a `SessionManagementDebugLog` boolean to allow debugging general session state.
- Made the sessions map private and add a lock, to prevent race conditions.
- Changed the behaviour of `GatheringFlags.DisconnectChangeOwner` flag. Only the owner is changed and the host is left untouched.
    - Fixes gatherings getting stuck.
- Changed the behaviour of `UpdateSessionURL` and `UpdateSessionHost`. Only the host is changed and the owner is left untouched.
- Send the `GatheringUnregistered` event when a gathering is deleted.

I will implement this change in the CTGP-7 server to investigate the effects. If they are positive will convert from a draft to a PR.


globals/matchmaking_utils.go
```diff
@@ -9,24 +10,53 @@ import (
...
 	notifications_types "github.com/PretendoNetwork/nex-protocols-go/v2/notifications/types"
 	"golang.org/x/exp/slices"
 )
 
+var sessions map[uint32]*CommonMatchmakeSession
+var sessionsMutex = sync.RWMutex{}
+
+var SessionManagementDebugLog = false
+
+func MakeSessions() {
```
DaniElectra, line 33: Can we add some godoc comment with a brief explanation of the function? Same for `GetSession` and `EachSession`

jonbarrow:
Is there a reason you didn't use `MutexMap` from `nex-go` here? We already provide a data structure for this kind of functionality

PabloMK7:
Because when I started implementing it I didn't know what I wanted yet. If stuff works it can be migrated.

PabloMK7:
Reviewed `MutexMap` and it doesn't really provide what's needed. It only fixes concurrent accesses to the map, but doesn't allow locking it for entire functions, which is what's needed here.

EDIT: It's actually possible to lock the mutexmap manually by calling `Lock` or `Unlock` on it, but then it's not possible to access its members without locking it again, causing deadlocks.

PabloMK7:
Ready for review after leaving the server running overnight and not noticing any issues nor stuck rooms in MK7.

globals/matchmaking_utils.go
```diff
@@ -131,21 +128,171 @@ func RemoveConnectionFromAllSessions(connection *nex.PRUDPConnection) {
...
+	// If the owner is the one being removed...
+	if ownerPID.Equals(connection.PID()) {
+		// * This flag tells the server to change the matchmake session owner if they disconnect
+		// * If the flag is not set, delete the session
+		// * More info: https://nintendo-wiki.pretendo.network/docs/nex/protocols/match-making/types#flags
+		if session.GameMatchmakeSession.Gathering.Flags.PAND(match_making.GatheringFlags.DisconnectChangeOwner) == 0 {
+			removeSessionImpl(connection, gathering)
+			return
+		} else {
+			// TODO: Figure out if this is the correct way to do it!
```
DaniElectra: I think we have determined this is correct?

globals/matchmaking_utils.go
```diff
@@ -131,21 +128,171 @@ func RemoveConnectionFromAllSessions(connection *nex.PRUDPConnection) {
...
+			changeSessionOwnerImpl(connection, gathering, true)
+		}
+	}
+	if !gracefully {
+		// Only send the disconnected notification if the connection disonnected ungracefully
+		endpoint := connection.Endpoint().(*nex.PRUDPEndPoint)
+		server := endpoint.Server
+
+		category := notifications.NotificationCategories.Participation
+		subtype := notifications.NotificationSubTypes.Participation.Disconnected
```
DaniElectra: We could inline the "graceful" notification on here, so that we don't have to handle it on `EndParticipation` and avoid race conditions

```suggestion
	var subtype uint32
	if graceful {
		subtype = notifications.NotificationSubTypes.Participation.Ended
	} else {
		subtype = notifications.NotificationSubTypes.Participation.Disconnected
	}
```

match-making/update_session_host.go
```diff
@@ -45,7 +51,7 @@ func (commonProtocol *CommonProtocol) updateSessionHost(err error, packet nex.Pa
 	}
 
 	originalOwner := session.GameMatchmakeSession.Gathering.OwnerPID
-	session.GameMatchmakeSession.Gathering.OwnerPID = connection.PID().Copy().(*types.PID)
+	session.GameMatchmakeSession.Gathering.OwnerPID = connection.PID().Copy().(*types.PID)
```
DaniElectra: There are two extra spaces at the end of this line

globals/matchmaking_utils.go
```diff
@@ -131,21 +128,171 @@ func RemoveConnectionFromAllSessions(connection *nex.PRUDPConnection) {
...
+		if gracefully {
+			grace = "gracefully"
+		} else {
+			grace = "ungracefully"
 		}
+		globals.Logger.Infof("GID %d: Removed PID %d %s", gathering, connection.PID().LegacyValue(), grace)
+	}
 
-		gid = FindConnectionSession(connection.ID)
+	// If there are no more participants remove the session
```
DaniElectra:
```suggestion
	// If there are no more participants, remove the session
```

globals/matchmaking_utils.go
```diff
@@ -197,7 +351,10 @@ func FindSessionByMatchmakeSession(pid *types.PID, searchMatchmakeSession *match
 	// TODO - This whole section assumes legacy clients. None of it will work on the Switch
 	var friendList []uint32
 	for _, sessionIndex := range candidateSessionIndexes {
-		sessionToCheck := Sessions[sessionIndex]
+		sessionToCheck, ok := sessions[sessionIndex]
+		if !ok {
+			continue
+		}
 		if sessionToCheck.ConnectionIDs.Size() >= int(sessionToCheck.GameMatchmakeSession.MaximumParticipants.Value) {
```
DaniElectra:
```suggestion
		}

		if sessionToCheck.ConnectionIDs.Size() >= int(sessionToCheck.GameMatchmakeSession.MaximumParticipants.Value) {
```

match-making/update_session_url.go
```diff
@@ -16,75 +13,20 @@ func (commonProtocol *CommonProtocol) updateSessionURL(err error, packet nex.Pac
...
-			messagePacket.SetSourceVirtualPortStreamType(target.StreamType)
-			messagePacket.SetSourceVirtualPortStreamID(endpoint.StreamID)
-			messagePacket.SetDestinationVirtualPortStreamType(target.StreamType)
-			messagePacket.SetDestinationVirtualPortStreamID(target.StreamID)
-			messagePacket.SetPayload(rmcRequestBytes)
-
-			server.Send(messagePacket)
-
-			return false
-		})
```
jonbarrow: While this may not be for changing the owner, @DaniElectra does not sending *anything* here seem correct to you? I scanned all the MK7 uploads we've gotten through user submissions and MK7 DOES use the "Host changed" notification type, which seems like it would be sent here?

globals/matchmaking_utils.go
```diff
@@ -131,21 +128,171 @@ func RemoveConnectionFromAllSessions(connection *nex.PRUDPConnection) {
...
+	lenParticipants := session.ConnectionIDs.Size()
+
+	if SessionManagementDebugLog {
+		var grace string
+		if gracefully {
+			grace = "gracefully"
+		} else {
+			grace = "ungracefully"
 		}
+		globals.Logger.Infof("GID %d: Removed PID %d %s", gathering, connection.PID().LegacyValue(), grace)
```
jonbarrow: This goes for all these kinds of logs, but using `LegacyValue` here isn't entirely safe. That method should ONLY be used when you NEED a uint32, and you can gurantee that the value is within that range. Otherwise `Value` should be used, in cases where the PID is larger than a uint32

globals/matchmaking_utils.go
```diff
@@ -9,24 +10,53 @@ import (
...
 // GetAvailableGatheringID returns a gathering ID which doesn't belong to any session
 // Returns 0 if no IDs are available (math.MaxUint32 has been reached)
 func GetAvailableGatheringID() uint32 {
 	return CurrentGatheringID.Next()
 }
 
-// FindOtherConnectionID searches a connection ID on the session that isn't the given one
-// Returns 0 if no connection ID could be found
-func FindOtherConnectionID(excludedConnectionID uint32, gatheringID uint32) uint32 {
+func findOtherConnectionIDImpl(excludedConnectionID uint32, gatheringID uint32) uint32 {
```
jonbarrow, line 91: I'm not a *huge* fan of the way these `Impl` functions are named/handled, though I understand why the implementation is separate from the exported function

This `Impl` convention could be refactored out by moving all of this out of the global scope and moving it to a struct, and having the functions currently named with `Impl` as just unexported methods

Something like

```go
package main

import "fmt"

type sessionsManager struct{}

func (sm *sessionsManager) FindOtherConnectionID() {
	fmt.Println("Exported")

	sm.findOtherConnectionID()
}

func (sm *sessionsManager) findOtherConnectionID() {
	fmt.Println("Unexported")
}

func main() {
	sm := &sessionsManager{}

	sm.FindOtherConnectionID()
	// Prints both Exported and Unexported
}
```

@DaniElectra thoughts?

globals/matchmaking_utils.go
jonbarrow: Just a nitpick to keep this consistent with other comments

```suggestion
			// TODO - Figure out if this is the correct way to do it!
```

globals/matchmaking_utils.go
jonbarrow: Nitpick. This goes for all comments. We use the syntax from [Better Comments](https://marketplace.visualstudio.com/items?itemName=aaron-bond.better-comments) for comments, with the exception of `TODO` which is structured as `TODO -` rather than `TODO:`

```suggestion
	// * If there are no more participants remove the session
```

globals/matchmaking_utils.go
```diff
@@ -231,11 +388,14 @@ func FindSessionByMatchmakeSession(pid *types.PID, searchMatchmakeSession *match
 
 // FindSessionsByMatchmakeSessionSearchCriterias finds a gathering that matches with the given search criteria
 func FindSessionsByMatchmakeSessionSearchCriterias(pid *types.PID, searchCriterias []*match_making_types.MatchmakeSessionSearchCriteria, gameSpecificChecks func(searchCriteria *match_making_types.MatchmakeSessionSearchCriteria, matchmakeSession *match_making_types.MatchmakeSession) bool) []*CommonMatchmakeSession {
-	candidateSessions := make([]*CommonMatchmakeSession, 0, len(Sessions))
+	sessionsMutex.RLock()
+	defer sessionsMutex.RUnlock()
+	
+	candidatesessions := make([]*CommonMatchmakeSession, 0, len(sessions))
```
jonbarrow: Why was this casing changed? The variable name is made of 2 words, "candidate" and "sessions"

matchmake-extension/get_simple_playing_session.go
```diff
@@ -28,8 +28,9 @@ func (commonProtocol *CommonProtocol) getSimplePlayingSession(err error, packet
 	}
 
 	simplePlayingSessions := make(map[string]*match_making_types.SimplePlayingSession)
-
-	for gatheringID, session := range common_globals.Sessions {
+	
+	var errRet *nex.Error = nil
+	common_globals.EachSession((func(gatheringID uint32, session *common_globals.CommonMatchmakeSession) bool {
```
jonbarrow: `EachSession` returns a bool if the loop exited early. In this case the only time it would exist early is if an error occurred. This can be used to remove the need to define the extra `errRet` variable

Such as:

```go
if common_globals.EachSession(func(gatheringID uint32, session *common_globals.CommonMatchmakeSession) bool {
	for _, pid := range listPID.Slice() {
		key := fmt.Sprintf("%d-%d", gatheringID, pid.Value())

		if simplePlayingSessions[key] == nil {
			connectedPIDs := make([]uint64, 0)

			session.ConnectionIDs.Each(func(_ int, connectionID uint32) bool {
				player := endpoint.FindConnectionByID(connectionID)
				if player == nil {
					common_globals.Logger.Warning("Player not found")
					return false
				}
				connectedPIDs = append(connectedPIDs, player.PID().Value())
				return false
			})

			if slices.Contains(connectedPIDs, pid.Value()) {
				attribute0, err := session.GameMatchmakeSession.Attributes.Get(0)
				if err != nil {
					return true
				}

				simplePlayingSessions[key] = match_making_types.NewSimplePlayingSession()
				simplePlayingSessions[key].PrincipalID = pid.Copy().(*types.PID)
				simplePlayingSessions[key].GatheringID = types.NewPrimitiveU32(gatheringID)
				simplePlayingSessions[key].GameMode = session.GameMatchmakeSession.GameMode.Copy().(*types.PrimitiveU32)
				simplePlayingSessions[key].Attribute0 = attribute0.Copy().(*types.PrimitiveU32)
			}
		}
	}

	return false
}) {
	nex.NewError(nex.ResultCodes.Core.InvalidArgument, "change_error")
}
```

This is already the pattern being introduced in the [friends server rewrite](https://github.com/PretendoNetwork/friends/blob/e2a630d1a761356ae3f360258fa021e519dca92c/nex/friends-3ds/sync_friend.go#L56-L76)

jonbarrow (changes requested):

match-making/update_session_url.go
DaniElectra: I wasn't aware that the server sends the "Host changed" notification, I had assumed that it was only for recent versions of NEX

Considering that this is the only method that the game uses for updating the host, yeah it should probably be added

globals/matchmaking_utils.go
PabloMK7: Waiting for DaniElectra response, it's also possible to not export the functions by just making them lowercase. But it may be better to have the sessions maps in the struct instead of being something global.

globals/matchmaking_utils.go
PabloMK7: I think the casing was not changed, it was like that already.

matchmake-extension/get_simple_playing_session.go
PabloMK7: Except that if a new error is needed to be introduced, the pattern will have to be changed again.

globals/matchmaking_utils.go
jonbarrow: > it's also possible to not export the functions by just making them lowercase

That's already how it's implemented, both in your PR and my suggestion? My comment was about exported vs unexported, it was on the naming conventions of the methods

It's clear you want to export a function which calls an implementation, without exposing the implementation itself. This is fine, I'm just not a fan of the way the functions themselves are named currently. So my suggestion was to bring them into a "manager" style struct, renaming the `Impl` functions to just have the same name as the exported one but be unexported

Though now that I think about it, why *is* the implementation functions separate now? I see elsewhere you call the `Impl` functions directly, but since both the exported and `Impl` functions have the same signatures why not just call the exported function directly and not have this separation?

> But it may be better to have the sessions maps in the struct instead of being something global.

My example was minimal to try and get the point across but yes ideally in this example everything to do with sessions would be contained in the manager (the lock, the map, etc)

matchmake-extension/get_simple_playing_session.go
jonbarrow: Correct. But that's not currently the case, and I don't see it needing to be the case any time soon, if ever? There's only ever one area where this snippet will error out and need to bail early, therefore only one kind of error needs to be handled. If something changes and more areas have the potential to throw errors then that will likely come from other refactors, which then would warrant a pattern change. But currently this can fit within the existing patterns

globals/matchmaking_utils.go
PabloMK7: The impl is needed to be able to call functions without locking the sessions lock again. I will move everything to a struct after I figure out the stuck gathering issues.

PabloMK7:
I just added the `HostChanged` notification to `UpdateSessionURL`, but I made it only send the notification to the old gathering host, and below is why.
Looking at the user dumps, I have made the following observations:
- Observation 1: After calling `UpdateSessionURL` an `OwnerChanged` notification SOMETIMES is being sent, and the new owner PID doesn't have to match the user who called the method.
- Observation 2: When an user receives a `HostChanged` notification, they ALWAYS call `EndParticipation` afterwards.

Here is what I think is happening:
1) At some point, the room host becomes unstable enough for the rest of the clients to agree on calling `UpdateSessionURL`. The server updates the session host and sends the `HostChanged` notification to the old host only.
2) The old host receives the `HostChanged` notification and leaves the gathering, calling `EndParticipation`.
3) The server receives the `EndParticipation`, and sends the `OwnerChanged` notification to the rest of the players if they were the owner.

This theory explains both observations:
- Observation 1: In the cases when host == owner, we see the `OwnerChanged` notification because the host was kicked and was the owner. In the cases when host != owner, we don't see any notifications because the owner didn't change and the `HostChanged` notification is not send to everyone.
- Observation 2: The `HostChanged` notification causes the host to leave the room, that's why it always comes just before `EndParticipation`.

jonbarrow:
That all sounds logical to me. Barring any new observations or issues with that implementation I think that's fine?

PabloMK7:
From my observations the NewParticipant notification is only sent to the owner in MK7. This has to be verified with other games (cc @shutterbug2000 )

PR #29, opened 2024-04-19
ticket-granting: Set the StructureVersion for NEX >3.5

ashquarky:
Required for games such as Minecraft: Wii U Edition. This was added last year in e6b3e0b8425a343e1f0b42c6288a1573412e6b13 but reverted right after in 01db10686f581902c1fa8fdf6e6fd8c00d23ee6c, with the note that this is handled in nex-go - but I can't find any of that code today or any of the commits related to it, so I'm not sure what happened.

Happy to re-file against nex-go if someone's willing to point me in the right direction there.

### DIFF
--- a/ticket-granting/login_ex.go
+++ b/ticket-granting/login_ex.go
@@ -64,6 +64,10 @@ func (commonProtocol *CommonProtocol) loginEx(err error, packet nex.PacketInterf
 		pConnectionData.SpecialProtocols = specialProtocols
 		pConnectionData.StationURLSpecialProtocols = commonProtocol.StationURLSpecialProtocols
 		pConnectionData.Time = types.NewDateTime(0).Now()
+
+		if endpoint.LibraryVersions().Main.GreaterOrEqual("v3.5.0") {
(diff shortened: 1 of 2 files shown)


ticket-granting/login_ex.go
DaniElectra, line 1: I assume this is a CRLF vs LF line endings difference?

DaniElectra (approved):
Nice catch!

Before the rewrite this was indeed handled by nex-go inside the `RVConnectionData` structure, but that has been discarded since then

jonbarrow:
I'll approve this as well but I'm also curious about the CRLF v LF question before merging

jonbarrow (approved):

ticket-granting/login_ex.go
ashquarky: Yeah, had to whip out the hex editor but it looks like my IDE converted CRLF to LF. I thought git showed that change but apparently not anymore...

DaniElectra: Alright, that is ok :+1:

PR #30, opened 2024-05-03
fix(matchmake-extension): Fix result type from AutoMatchmakeWithSearchCriteria_Postpone to be AnyDataHolder

ashquarky:
A bit like the contentWritable bug, here a struct was created but then not actually passed to the writer.

[As per the wiki,](https://nintendo-wiki.pretendo.network/docs/nex/protocols/matchmake-extension/#15-automatchmakewithsearchcriteria_postpone) this should be a Data, not a raw MatchmakeSession.


jonbarrow:
Nice catch!

PR #31, opened 2024-05-03
feat!(datastore): Add requesting PID to GetObjectInfosByDataStoreSearchParam

ashquarky:
This is required for PUYOPUYOTETRIS, which doesn't use the perfectly good ownerPID field and instead sets a search target to mean "your own account". Very cool.

Breaking API change, but I don't think there are actually any consumers of the v2 DataStore API anyway?


jonbarrow:
LGTM

PR #34, opened 2024-06-24
feat!(ranking): Add new methods for the different ranking modes on GetRanking

ashquarky:
There's 5 different types of GetRanking call, asking for:
- global leaderboards
- global leaderboards, nearby to your own rank
- only your friends
- only your friends, nearby to your own rank
- your own ranking only

These are pretty fundamentally different requests that would need different db queries, etc. at the implementation level. Notably all but the first need knowledge of the PID of the user making the request (to get their rank and their friends).

### Changes:

Add new methods to the protocol for each of these use cases.

Submitting as a draft because I *know* `if rankingMode.Value == 0 {` is the wrong way to do it, but I couldn't find any other protocol with constants like that for me to reference against. If someone count point out where I should define constants for this sort of thing I'll happily clean up the magic numbers.

There's a highly WIP implementation of this API here: https://github.com/ashquarky/puyo-wiiu/blob/ranking/datastore/puyo_ranking.go

Note the dramatically different database requests needed for each mode.

### DIFF
--- a/ranking/get_ranking.go
+++ b/ranking/get_ranking.go
@@ -10,31 +10,6 @@ import (
 )
 
 func (commonProtocol *CommonProtocol) getRanking(err error, packet nex.PacketInterface, callID uint32, rankingMode types.UInt8, category types.UInt32, orderParam ranking_types.RankingOrderParam, uniqueID types.UInt64, principalID types.PID) (*nex.RMCMessage, *nex.Error) {
-	if commonProtocol.GetRankingsAndCountByCategoryAndRankingOrderParam == nil {
-		common_globals.Logger.Warning("Ranking::GetRanking missing GetRankingsAndCountByCategoryAndRankingOrderParam!")
-		return nil, nex.NewError(nex.ResultCodes.Core.NotImplemented, "change_error")
-	}
-
-	if commonProtocol.GetNearbyRankingsAndCountByCategoryAndRankingOrderParam == nil {
-		common_globals.Logger.Warning("Ranking::GetRanking missing GetNearbyRankingsAndCountByCategoryAndRankingOrderParam!")
-		return nil, nex.NewError(nex.ResultCodes.Core.NotImplemented, "change_error")
-	}
-
-	if commonProtocol.GetFriendsRankingsAndCountByCategoryAndRankingOrderParam == nil {
-		common_globals.Logger.Warning("Ranking::GetRanking missing GetFriendsRankingsAndCountByCategoryAndRankingOrderParam!")
-		return nil, nex.NewError(nex.ResultCodes.Core.NotImplemented, "change_error")
-	}
-
-	if commonProtocol.GetNearbyFriendsRankingsAndCountByCategoryAndRankingOrderParam == nil {
-		common_globals.Logger.Warning("Ranking::GetRanking missing GetNearbyFriendsRankingsAndCountByCategoryAndRankingOrderParam!")
-		return nil, nex.NewError(nex.ResultCodes.Core.NotImplemented, "change_error")
-	}
-
-	if commonProtocol.GetOwnRankingByCategoryAndRankingOrderParam == nil {
-		common_globals.Logger.Warning("Ranking::GetRanking missing GetOwnRankingByCategoryAndRankingOrderParam!")
-		return nil, nex.NewError(nex.ResultCodes.Core.NotImplemented, "change_error")
-	}
-
 	if err != nil {
 		common_globals.Logger.Error(err.Error())
 		return nil, nex.NewError(nex.ResultCodes.Ranking.InvalidArgument, "change_error")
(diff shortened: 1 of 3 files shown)


DaniElectra:
We define the relevant constants for protocols inside nex-protocols-go. See this example: https://github.com/PretendoNetwork/nex-protocols-go/blob/master/nintendo-notifications/notification_types.go

We could also replace that big `if` statement with a `switch` case

ranking/get_ranking.go
```diff
@@ -14,15 +14,54 @@ func (commonProtocol *CommonProtocol) getRanking(err error, packet nex.PacketInt
...
+	callerPid := principalID
+	// * PUYOPUYOTETRIS seems to do this sometimes
+	if callerPid.Value() == 0 {
+		callerPid = connection.PID()
+	}
+
+	var rankDataList *types.List[*ranking_types.RankingRankData]
+	var totalCount uint32
+
+	if rankingMode.Value == 0 {
```
JPeisach:
```suggestion
	switch rankingMode.Value {
	case 0:
	    rankDataList, totalCount, err = commonProtocol.GetRankingsAndCountByCategoryAndRankingOrderParam(category, orderParam)
	case 1:
		rankDataList, totalCount, err = commonProtocol.GetNearbyRankingsAndCountByCategoryAndRankingOrderParam(callerPid, category, orderParam)
	case 2:
		rankDataList, totalCount, err = commonProtocol.GetFriendsRankingsAndCountByCategoryAndRankingOrderParam(callerPid, category, orderParam)
	case 3:
		rankDataList, totalCount, err = commonProtocol.GetNearbyFriendsRankingsAndCountByCategoryAndRankingOrderParam(callerPid, category, orderParam)
	default: // rankingMode.Value == 4
		rankDataList, totalCount, err = commonProtocol.GetOwnRankingByCategoryAndRankingOrderParam(callerPid, category, orderParam)
	}
```
man I still cant figure out suggestions with GitHub for multiple lines

ranking/get_ranking.go
```diff
@@ -14,15 +15,56 @@ func (commonProtocol *CommonProtocol) getRanking(err error, packet nex.PacketInt
...
+	}
+
 	if err != nil {
 		common_globals.Logger.Error(err.Error())
 		return nil, nex.NewError(nex.ResultCodes.Ranking.InvalidArgument, "change_error")
 	}
 
 	connection := packet.Sender()
 	endpoint := connection.Endpoint()
+	callerPid := principalID
```
DaniElectra:
```suggestion
	callerPID := principalID
```

ranking/get_ranking.go
```diff
@@ -14,15 +15,56 @@ func (commonProtocol *CommonProtocol) getRanking(err error, packet nex.PacketInt
...
+
+	if commonProtocol.GetNearbyFriendsRankingsAndCountByCategoryAndRankingOrderParam == nil {
+		common_globals.Logger.Warning("Ranking::GetRanking missing GetNearbyFriendsRankingsAndCountByCategoryAndRankingOrderParam!")
+		return nil, nex.NewError(nex.ResultCodes.Core.NotImplemented, "change_error")
+	}
+
+	if commonProtocol.GetOwnRankingByCategoryAndRankingOrderParam == nil {
+		common_globals.Logger.Warning("Ranking::GetRanking missing GetOwnRankingByCategoryAndRankingOrderParam!")
+		return nil, nex.NewError(nex.ResultCodes.Core.NotImplemented, "change_error")
+	}
```
DaniElectra: I think we should redesign this so that server implementers don't have to implement *every* function in here, since games won't use every ranking mode. For this we can move these checks inside the matching switch case

ranking/get_ranking.go
```diff
@@ -5,24 +5,66 @@ import (
...
+		rankDataList, totalCount, err = commonProtocol.GetFriendsRankingsAndCountByCategoryAndRankingOrderParam(callerPID, category, orderParam)
+	case constants.RankingModeFriendNear:
+		if commonProtocol.GetNearbyFriendsRankingsAndCountByCategoryAndRankingOrderParam == nil {
+			common_globals.Logger.Warning("Ranking::GetRanking missing GetNearbyFriendsRankingsAndCountByCategoryAndRankingOrderParam!")
+			return nil, nex.NewError(nex.ResultCodes.Core.NotImplemented, "change_error")
+		}
+
+		rankDataList, totalCount, err = commonProtocol.GetNearbyFriendsRankingsAndCountByCategoryAndRankingOrderParam(callerPID, category, orderParam)
+	case constants.RankingModeUser:
+	default:
```
DaniElectra: Sorry didn't notice earlier, two things here:
1. This won't actually work the way you'd think since Go breaks out of switch cases by default, so nothing will happen in the case of `constants.RankingModeUser`. To fix this you'd want to add a `fallthrough` to make it work the same way as it would in C
2. Do we want to assume `constants.RankingModeUser` by default, is this correct? If it's just a guess, then I believe we should return InvalidArgument in the default case

DaniElectra (approved):
LGTM, seems trivial enough to directly merge

ashquarky:
Ta! :blue_heart:

PR #42, opened 2024-08-12
feat: Add extension point for CanJoinMatchmakeSession

ashquarky:
### Changes:

Adds the option for games to override CanJoinMatchmakeSession with their own version if needed. This is useful for, at least:

- Minecraft: Wii U Edition, which doesn't use NEX ParticipationPolicy at all and instead uses a combination of search attributes and application buffers to determine join eligibility. This is I guess done to allow "Friends of friends" matchmaking on the client; but we'd like to check it serverside too for security reasons.
- Splatoon's Splatfests, which involve matchmaking the second team into a room with `OpenParticipation` false - the current check forbids this. Presumably this is to prevent random individuals matching into a room without a whole team alongside them, which would make the room unplayable (any legit teams would find there aren't enough vacant participants)

I have a [sample of what Minecraft's matchmaking looks like here](https://github.com/PretendoNetwork/minecraft-wiiu/commit/fd75619953903f3cc09520b1d1db67e69e20e3b9). I'm not too sure if you like the design here, since it encourages reaching into the database directly (as I did in the minecraft diff and `database.FindGatheringByID` to get the match participants).

### DIFF
--- a/matchmake-extension/join_matchmake_session.go
+++ b/matchmake-extension/join_matchmake_session.go
@@ -38,7 +38,12 @@ func (commonProtocol *CommonProtocol) joinMatchmakeSession(err error, packet nex
 		return nil, nex.NewError(nex.ResultCodes.RendezVous.PermissionDenied, "change_error")
 	}
 
-	nexError = common_globals.CanJoinMatchmakeSession(commonProtocol.manager, connection.PID(), joinedMatchmakeSession)
+	// * Allow game servers to do their own permissions checks
+	if commonProtocol.CanJoinMatchmakeSession != nil {
+		nexError = commonProtocol.CanJoinMatchmakeSession(commonProtocol.manager, connection.PID(), joinedMatchmakeSession)
+	} else {
+		nexError = common_globals.CanJoinMatchmakeSession(commonProtocol.manager, connection.PID(), joinedMatchmakeSession)
+	}
 	if nexError != nil {
 		commonProtocol.manager.Mutex.Unlock()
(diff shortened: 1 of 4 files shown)


globals/matchmaking_globals.go
```diff
@@ -15,6 +16,7 @@ type MatchmakingManager struct {
 	Mutex                    *sync.RWMutex
 	GetUserFriendPIDs        func(pid uint32) []uint32
 	GetDetailedGatheringByID func(manager *MatchmakingManager, gatheringID uint32) (types.RVType, string, *nex.Error)
+	CanJoinMatchmakeSession  func(manager *MatchmakingManager, pid *types.PID, matchmakeSession *match_making_types.MatchmakeSession) *nex.Error
```
DaniElectra: This override should be added on the MatchmakeExtension common protocol, not here. The `MatchmakingManager` should not include any NEX-specific overrides (`match-making` and `match-making-ext` come from QRV, whereas `matchmake-extension` and the `MatchmakeSession` are specific to NEX).

The override should be done in the places where `CanJoinMatchmakeSession` is called, like for the other overrides present in the MatchmakeExtension common protocol

DaniElectra (commented):
The Minecraft use case seems fair enough.

The Splatfest issue sounds like something may not implemented correctly, like the `gidForParticipationCheck` which currently isn't used. Thus, the override probably shouldn't be required there (unless proven otherwise)

DaniElectra (approved):
LGTM

PR #45, opened 2024-11-17
feat: Initial support for persistent gatherings/communities

DaniElectra:
Resolves #XXX

### Changes:

Implement methods that are needed for Mario Kart 7 communities to work. Note that support for communities on MK7 is partial since the community statistics don't load because legacy Ranking isn't implemented. Aside from that, players can create communities and join others without issues. Other games which use persistent gatherings may or may not work.

In order to support the `ParticipationCount`, we replace matchmake session joins with a wrapper which checks if the session is attached to a community, and if it is, it will increment the participation count of the player in a new table named `community_participations`. The `MatchmakeSessionCount` is handled more easily by checking the sessions that belong to the corresponding community.

A new parameter is also added named `PersistentGatheringCreationMax` with a default value of 4, as reported and tested on various games. This allows game servers to change the maximum number of active persistent gatherings that a player can create. For example, Mario Kart 7 supports up to 8 persistent gatherings instead of the default of 4.

In Mario Kart 7 there is no limitation on the number of players that can "join" to a community. That is because they don't really join to it but they create matchmake sessions linked to the persistent gathering (in fact, the `MaximumParticipants` parameter on persistent gatherings is set to 0). Thus, the `participants` parameter is unused in communities (at least on MK7) and we instead log community participations with a new tracking table `tracking.participate_community`.

Some changes also had to be done in other places like participant disconnection handling or gathering registrations in order to implement persistent gatherings accurately.


globals/utils.go
```diff
@@ -77,6 +77,36 @@ func CheckValidMatchmakeSession(matchmakeSession match_making_types.MatchmakeSes
...
+	}
+
+	// * Only allow normal and password-protected community types
+	if persistentGathering.CommunityType != 0 && persistentGathering.CommunityType != 1 {
+		return false
+	}
+
+	// * All strings must have a length lower than 256
+	//
+	// TODO - Can the password actually be up to 256 characters?
```
jonbarrow: No actually. By default the matchmaking passwords can only be 32 characters for user passwords and 16 characters for system passwords

https://github.com/tech-ticks/RTDXTools/blob/232a7797e01369e9c12704f58fdde65dd3ac1c32/Assets/Scripts/Stubs/Generated/Assembly-CSharp/NexPlugin/Matchmake.cs#L27-L28

globals/utils.go
```diff
@@ -77,6 +77,36 @@ func CheckValidMatchmakeSession(matchmakeSession match_making_types.MatchmakeSes
...
+// CheckValidPersistentGathering checks if a PersistentGathering is valid
+func CheckValidPersistentGathering(persistentGathering match_making_types.PersistentGathering) bool {
+	if !CheckValidGathering(persistentGathering.Gathering) {
+		return false
+	}
+
+	// * Only allow normal and password-protected community types
+	if persistentGathering.CommunityType != 0 && persistentGathering.CommunityType != 1 {
+		return false
+	}
```
jonbarrow: We should probably define these constants somewhere, which can get rid of this comment and make this a bit more clear. We know the real names of these types now

https://github.com/tech-ticks/RTDXTools/blob/232a7797e01369e9c12704f58fdde65dd3ac1c32/Assets/Scripts/Stubs/Generated/Assembly-CSharp/NexPlugin/Matchmake.cs#L113-L118

match-making/database/register_gathering.go
```diff
@@ -38,8 +38,8 @@ func RegisterGathering(manager *common_globals.MatchmakingManager, pid types.PID
 		$10,
 		$11
 	) RETURNING id`,
-		pid,
-		pid,
+		uint64(ownerPID),
+		uint64(hostPID),
```
jonbarrow: Are these casts required? Part of the goal with the types rewrite is to allow code that doesn't require so many casts. `PID` for example can be used like any other number through normal means, is Postgres treating this differently?

```go
package main

import "fmt"

type PID uint64

func main() {
	pid := 100
	total := pid + 50

	fmt.Println(total) // 150
}
```

This question applies everywhere these casts are seen (only doing one comment)

match-making/database/register_gathering.go
```diff
@@ -8,8 +8,8 @@ import (
 	match_making_types "github.com/PretendoNetwork/nex-protocols-go/v2/match-making/types"
 )
 
-// RegisterGathering registers a new gathering on the databse. No participants are added
-func RegisterGathering(manager *common_globals.MatchmakingManager, pid types.PID, gathering *match_making_types.Gathering, gatheringType string) (types.DateTime, *nex.Error) {
+// RegisterGathering registers a new gathering on the database. No participants are added
+func RegisterGathering(manager *common_globals.MatchmakingManager, ownerPID types.PID, hostPID types.PID, gathering *match_making_types.Gathering, gatheringType string) (types.DateTime, *nex.Error) {
```
jonbarrow, line 12: If we wanted to we can condense this signature since Go allows you to define multiple arguments of the same type right after each other:

```go
package main

import "fmt"

func main() {
	test(1, "hello", "world", 2)
}

func test(arg1 uint8, arg2, arg3 string, arg4 uint64) {
	fmt.Println(arg1) // 1
	fmt.Println(arg2) // hello
	fmt.Println(arg3) // world
	fmt.Println(arg4) // 2
}
```

We don't have to do this though, up to you. I don't think we use this much elsewhere rn anyway, if at all?

matchmake-extension/create_community.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	// TODO - Is this right? Mario Kart 7 sets 0 max participants
+	if community.MaximumParticipants > 0 {
+		_, nexError = match_making_database.JoinGathering(commonProtocol.manager, uint32(community.ID), connection, 1, string(strMessage))
+		if nexError != nil {
+			commonProtocol.manager.Mutex.Unlock()
+			return nil, nexError
+		}
+	}
```
jonbarrow: Iirc "0" would mean "no limit". Typically in NEX whenever "0" is set it means disabled/invalid/unset, so in this case I would interpret it as there being no max amount of participants (which makes sense, seeing as persistent gatherings like these could have any number of players)

matchmake-extension/database/get_detailed_gathering_by_id.go
```diff
@@ -18,19 +18,43 @@ func GetDetailedGatheringByID(manager *common_globals.MatchmakingManager, gather
...
+
+		matchmakeSessionCount, nexError := GetPersistentGatheringSessionCount(manager, gatheringID)
+		if nexError != nil {
+			return nil, "", nexError
+		}
+
+		participationCount, nexError := GetPersistentGatheringParticipationCount(manager, gatheringID, sourcePID)
+		if nexError != nil {
+			return nil, "", nexError
+		}
```
jonbarrow: Doing this in 3 separate calls *could* lead to race conditions, and it hits the DB with several calls. I get the feeling this can be reduced down to a single call, since all 3 of these are just getting different stats for the same gathering?

I know some of these functions are reused elsewhere, but it may be worth just duplicating the code a few times for specific use cases to get optimal database performance and remove race conditions?

matchmake-extension/database/get_official_communities.go
```diff
@@ -0,0 +1,137 @@
...
+		if nexError != nil {
+			common_globals.Logger.Critical(nexError.Error())
+			continue
+		}
+
+		resultParticipationCount, nexError := GetPersistentGatheringParticipationCount(manager, uint32(resultPersistentGathering.ID), uint64(sourcePID))
+		if nexError != nil {
+			common_globals.Logger.Critical(nexError.Error())
+			continue
+		}
```
jonbarrow: Same comment as before about reducing the DB calls (our DB gets pretty hammered as is). I won't leave a new comment for each instance, only doing the 2

matchmake-extension/database/join_matchmake_session.go
```diff
@@ -0,0 +1,41 @@
...
+
+// JoinMatchmakeSession joins participants from the same connection into a MatchmakeSession. Returns the new number of participants
+func JoinMatchmakeSession(manager *common_globals.MatchmakingManager, matchmakeSession match_making_types.MatchmakeSession, connection *nex.PRUDPConnection, vacantParticipants uint16, joinMessage string) (uint32, *nex.Error) {
+	newParticipants, nexError := match_making_database.JoinGathering(manager, uint32(matchmakeSession.ID), connection, vacantParticipants, joinMessage)
+	if nexError != nil {
+		return 0, nexError
+	}
+
+	// TODO - Should we return the error in these cases?
+	if matchmakeSession.MatchmakeSystemType == 5 { // * Attached to a persistent gathering
```
jonbarrow: Same comment about the constants, we know the real names of these https://github.com/tech-ticks/RTDXTools/blob/232a7797e01369e9c12704f58fdde65dd3ac1c32/Assets/Scripts/Stubs/Generated/Assembly-CSharp/NexPlugin/Matchmake.cs#L48-L52

jonbarrow (commented):
Overall looks pretty good! Just a couple notes

match-making/database/register_gathering.go
DaniElectra: I had thought that it wouldn't, but looking on the `database/sql` source code looks like it may work? https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/database/sql/convert.go;drc=2a7ca156b8189c68c0a29b4c66194a42c5ce3c9b;l=430

Will have to test it properly

globals/utils.go
DaniElectra: Those are for matchmake session passwords, not persistent gathering. Though I guess I can extrapolate the user passwords value and maybe add a TODO for actually checking it is correct

match-making/database/register_gathering.go
DaniElectra: I'm not sure how commonly we follow this, I don't think we do it anywhere else yeah

matchmake-extension/database/get_detailed_gathering_by_id.go
DaniElectra: There are no race conditions since all matchmaking functions are executed within a mutex. Though it makes sense to optimize it, but I would need some help to craft the inlined queries

match-making/database/register_gathering.go
jonbarrow: Do you have an opinion on this sort of styling? I don't have a strong one. I just checked the protocols lib and we don't follow it there. For example:

```go
ModifyCurrentGameAttribute func(err error, packet nex.PacketInterface, callID uint32, gid types.UInt32, attribIndex types.UInt32, newValue types.UInt32) (*nex.RMCMessage, *nex.Error)
```

However that's because these files were basically all generated automatically, and that was just the easiest way to do that

matchmake-extension/database/get_detailed_gathering_by_id.go
jonbarrow: Might have to call in @binaryoverload for some help too, as I'm by no means a postgres expert

matchmake-extension/create_community.go
DaniElectra: That could be the case, but it would mean that all MK7 communities would have one "participant" under the current code, since you don't actually join communities as explained in the PR description, you only create matchmake sessions attached to a community

match-making/database/register_gathering.go
DaniElectra: I think it should remain as-is on nex-protocols-go as it makes more clear each parameter type. I don't have preferences in other places though

matchmake-extension/create_community.go
TraceEntertains: Smash 4 seems to set a MaximumParticipants count over 0, so I agree with Jon

matchmake-extension/database/get_detailed_gathering_by_id.go
binaryoverload:
```sql
SELECT community_type,
       password,
       attribs,
       application_buffer,
       participation_start_date,
       participation_end_date,
       (SELECT COUNT(ms.id)
        FROM matchmaking.matchmake_sessions AS ms
                 INNER JOIN matchmaking.gatherings AS g ON ms.id = g.id
        WHERE g.registered = true
          AND ms.matchmake_system_type = 5
          AND ms.attribs[1] = $1) AS persistent_gathering_session_count
       (SELECT cp.participation_count
        FROM matchmaking.community_participations AS cp
        WHERE cp.user_pid = $1
          AND cp.gathering_id = $2) as persistent_gathering_participation_count
FROM mh4u.matchmaking.persistent_gatherings
WHERE id = $1;
```

We can use inline queries like this - allowing the two counts to be set in the same query 

This could be all put into the `GetPersistentGatheringByGathering` function

I don't have a way to test this query, as the running servers don't have the community_participations table

matchmake-extension/create_community.go
jonbarrow: > but it would mean that all MK7 communities would have one "participant" under the current code, since you don't actually join communities as explained in the PR description, you only create matchmake sessions attached to a community

This is correct behavior. I just verified this from a NN packet dump. In the dump the following happens, in order:

- `MatchmakeExtension->CreateCommunity` is called
  - The `PersistentGathering` only has the `Persistent` flag (0x1) set, all other flags are unset
  - Min and max participants are set to 0 (interpreted as "unlimited", which makes sense for a community)
- `MatchmakeExtension->CreateCommunity` succeeds and returns the GID `50378197`
- `MatchmakeExtension->FindCommunityByParticipant` is called with the PID `1873532003` (the current user)
- `MatchmakeExtension->FindCommunityByParticipant` succeeds and returns the `PersistentGathering` with GID `50378197` that was made before
  - `m_ParticipationCount` of the `PersistentGathering` is already set to 1
  
Having the creator join  immediately is intended behavior. We know that users can be in multiple gatherings at once now, and we also know that being registered as a participant does not inherently mean you're in a session, so this is fine

I'd be willing to bet that the flag `AllowNoParticipant` (0x80), which allows a persistent gathering to exist with 0 participants, was not added until a later protocol version so adding a user (the creator) to the persistent gathering was the only way to keep it alive even when no one was playing it as a "hack"

DaniElectra: I see, sounds good then

TraceEntertains:
If possible, could some research be done into CreateOrJoinMatchmakeSession? It's used in Smash 4 and I think it may be a common impl, would like to see it in here now if possible instead of needing another PR later lol

jonbarrow:
> If possible, could some research be done into CreateOrJoinMatchmakeSession? It's used in Smash 4 and I think it may be a common impl, would like to see it in here now if possible instead of needing another PR later lol

This PR is aimed at persistent gatherings. To keep in line with our [pull request guidelines](https://github.com/PretendoNetwork/.github/blob/master/.github/CONTRIBUTING.md#pull-requests), unless `CreateOrJoinMatchmakeSession` is used for persistent gatherings it should be handled in a different pull request

We intentionally try to keep pull request scopes as small as possible

matchmake-extension/create_community.go
jonbarrow: > I see, sounds good then

In the dump I have both `m_ParticipationStartDate` and `m_ParticipationEndDate` are set to 0 as well. I've found references to a "never" value in Xenoblade and WATCH_DOGS, with WATCH_DOGS actually having some code that uses it (specifically a "isNever" method, which seems to check if the input `DateTime` is 0?)

I think we can also safely assume that a value of 0 for a DateTime means "never", and has the same "unlimited" meaning as the other 0 values here (in this case that means that the min and max participants have 0 max values, there is no start date for when users can join, and there is no end date for blocking new joins)

Though this only seems to apply to matchmaking. In DataStore a different date seems to be used for this?

TraceEntertains:
In the context of smash 4, createorjoinmatchmakesession is used in a persistentgathering

Generally matchmakesession is used heavily in regards to persistentgatherings, as you can see in the contents of the PR already lol

matchmake-extension/create_community.go
DaniElectra: I see. There is also a reference of a DateTime labeled as "future" on NintendoClients: https://github.com/kinnay/NintendoClients/blob/e2b673bac6d7781e83f83ae2c5d0c34a2092d72e/nintendo/nex/common.py#L246

I'm not sure what it represents though

jonbarrow:
> In the context of smash 4, createorjoinmatchmakesession is used in a persistentgathering
> 
> Generally matchmakesession is used heavily in regards to persistentgatherings, as you can see in the contents of the PR already lol

By virtue of being a part of matchmaking, yes sessions are used. But that does not inherently mean that they are critical to persistent gatherings/communities since those exist *outside* of active sessions

That being said, `JoinOrCreateMatchmakeSession` is NOT a common method. It is specific to Smash4, and so it should not be here regardless. All game-specific functionality goes in that games server

TraceEntertains:
Oh, fair enough lol
Was rushing out the door so couldn't check, guess I thought it was a common method?

matchmake-extension/create_community.go
jonbarrow: > I see. There is also a reference of a DateTime labeled as "future" on NintendoClients: [kinnay/NintendoClients@`e2b673b`/nintendo/nex/common.py#L246](https://github.com/kinnay/NintendoClients/blob/e2b673bac6d7781e83f83ae2c5d0c34a2092d72e/nintendo/nex/common.py#L246)
> 
> I'm not sure what it represents though

This looks to be what I've seen in DataStore (the year 9999 jumps out to me). Dates like this seem to be used for things like expiration times, likely to prevent an object from ever expiring? It seems to use the highest possible value for the type, and in the context of DataStore these are usually called "invalid" types (IE, `INVALID_DATA_TYPE` and `INVALID_PERSISTENCE_SLOT_ID` both being 65535). When these values are used, the field is ignored, so that lines up (for example if `INVALID_PERSISTENCE_SLOT_ID` is used then an object is not put into a persistent slot)

jonbarrow:
It definitely has a name that makes it sound common, yeah. But it's only in [Smash4](https://nintendo-wiki.pretendo.network/docs/nex/protocols/matchmake-extension/super-smash-bros-4#41-joinorcreatematchmakesession). It replaces method [`MatchmakeExtension::FindMatchmakeSessionByGatheringIdDetail`](https://nintendo-wiki.pretendo.network/docs/nex/protocols/matchmake-extension/#41-findmatchmakesessionbygatheringiddetail) as they share the same ID

matchmake-extension/create_community.go
TraceEntertains: It would definitely be nice to have some kind of implementation for future, never, and maybe the go zero date (0001-01-01), as a datetime of 0 returns a negative year. Needed future and the zero date in datastore for database timestamp stuff

jonbarrow: > as a datetime of 0 returns a negative year

Can you provide a sample that shows this? Like I mentioned earlier the times used by MK7 communities are the value 0, which in NEX Viewer comes out to `Wed Nov 29 1899 19:00:00 GMT-0500` (in  EST, at least)

The "never" value is almost certainly just the value 0, given that MK7 uses 0 to represent an unlimited time frame and WATCH_DOGS `DateTime` method `isNever` seems to compare the value to 0 (internally the value is stored kinda funky so the method does some extra stuff, but ultimately it seems to boil down to a 0 check)

TraceEntertains: Today I learned that github auto refreshes comments but not conversation comments

I'm not home rn so I can't give an exact example, but when you convert a DateTime of 0 to a go timestamp, it becomes something like -0001-11-30 and I'm not sure if postgres is really a fan of that

jonbarrow: That sounds like a bug on our end, then. A negative date should never happen

DaniElectra: FWIW yeah the date seems to be converting to a negative value. Though Postgres seems to be able to handle it, when querying for fields with time 0 it shows `0002-11-30 00:00:00 BC`

TraceEntertains: I don't think it's a bug on our end really, just a quirk of the go implementation we didn't account for

It creates a time object with everything 0, which go turns into a negative date

TraceEntertains:
Looks good to me

jonbarrow:
@DaniElectra I've made a new release of the protocols lib, which includes all the new constants we've added recently. We should probably update this PR to target that, and use the constants where possible, then I think this is ready to merge?

DaniElectra:
I have updated the go modules, and I believe the constants are used where possible

PR #49, opened 2025-01-27
feat(matchmake-extension): Implement NotificationData methods

DaniElectra:
Resolves #XXX

### Changes:

**NOTE: The changes currently haven't been tested, so marking as draft**

The NotificationData methods are used by games to send notifications to friends about user activity, among others. These notifications are created or updated using `UpdateNotificationData`, which the server will register and send to the connected friends (as seen on Mario Tennis Open).

The lifetime of these notifications is the same as the connection of the user who sends them. That is, when a user sends a notification and then disconnects, the notifications will be discarded.

All notifications sent over `UpdateNotificationData` are logged inside the `tracking.notification_data` table to prevent abuse. The type of these notifications is also constrained to a range of specific values reserved for game-specific purposes (from 101 to 108).


DaniElectra:
This has been tested on various games and working without issues after fixing some bugs, so merging

PR #52, opened 2025-03-15
feat: Implement token validation on TicketGranting and SecureConnection

DaniElectra:
Resolves #XXX

### Changes:

Add callback function `ValidateLoginData` on the `ticket-granting` and `secure-connection` common protocols to allow validation of the given login data on the `LoginEx` or `RegisterEx` methods. Alongside that, add a toggle on the `secure-connection` common protocol for enabling or disabling the insecure `Register` method, like the one for `Login` on `ticket-granting`.

Also implement the Pretendo token validation inside the globals section, so that both common protocols can use them. The common protocols include an option to use the Pretendo token validation `SetPretendoValidation` on that protocol, but custom handling can be added by setting the callback function to a custom one.

And finally, replace the error handling on the protocols to set any errors that aren't related to invalid arguments on `%retval%`. This is accurate as per previous research, since this is where errors like maintenace or validation failures get populated.

By default, both of the insecure methods are disabled. That means that current game servers that use the insecure `Register` method will have to enable them using `EnableInsecureRegister`. Server implementers **MUST NOT** enable both of the insecure methods at the same time.

The new `SetPretendoValidation` function expects an AES key as an argument. This matches with the AES key that is stored on the server list from the account server, after being hex-decoded. Servers are expected to set the AES key from an environment variable and validate that the input provided is valid (the key must be decoded using `hex.DecodeString` before being provided to the validation fucntion).


globals/authentication.go
```diff
@@ -0,0 +1,139 @@
...
+	// Decrypt the token body
+	block, err := aes.NewCipher(aesKey)
+	if err != nil {
+		return nil, nex.NewError(nex.ResultCodes.Authentication.ValidationFailed, err.Error())
+	}
+
+	expectedChecksum := binary.BigEndian.Uint32(encryptedToken[0:4])
+	encryptedBody := encryptedToken[4:]
+
+	decrypted := make([]byte, len(encryptedBody))
```
jonbarrow, line 113: This is missing a check against `aes.BlockSize`. If a token with an invalid block size is given, then this function will panic. Something like:

```go
if len(encryptedBody)%aes.BlockSize != 0 {
    return nil, return nil, nex.NewError(nex.ResultCodes.Authentication.ValidationFailed, fmt.Sprintf("whatever error"))
}
```

jonbarrow (changes requested):

MiniBioOne:
Hi! I'm working on my own NEX server for Fast Racing NEO (https://github.com/MiniBioOne/fast-racing-neo) and I'm confused on what exactly I need to add to my server to prepare for this PR. Just the environment variable (and code related to it)? And do I need to add any of those insecure login functions?

globals/authentication.go
```diff
@@ -0,0 +1,143 @@
...
+	"time"
+
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	account_management_types "github.com/PretendoNetwork/nex-protocols-go/v2/account-management/types"
+	ticket_granting_types "github.com/PretendoNetwork/nex-protocols-go/v2/ticket-granting/types"
+)
+
+// ValidatePretendoLoginData validates the given Pretendo login data
+func ValidatePretendoLoginData(pid types.PID, loginData types.DataHolder, aesKey []byte) *nex.Error {
```
jonbarrow, line 21: What are your thoughts on changing the return signature here to return the token as well as the error? We can ignore the token being returned in `SetPretendoValidation` since we only need the error there and the token structure is specific to us, but having this function return the token as well once validated would make it so we can piggy back off this function inside `NintendoCreateAccount` without having to process the incoming token a 2nd time. Currently the way this would work is validating the token using `ValidatePretendoLoginData` inside `NintendoCreateAccount`, then processing the incoming data *again* to get the token value and decrypt it for use in the method

jonbarrow: Or maybe `ValidateLoginData` should also be updated? That way we can make `NintendoCreateAccount` *also* use the generic `ValidateLoginData` function instead of ONLY calling our Pretendo-specific one?

DaniElectra: I'm not against it, but I don't see why it's useful in that case? The only thing being used from the token outside validation is the PID, but you can also get it from the PRUDP connection, and that one is being validated here

As an additional note, the `NintendoCreateAccount` method uses an exclusive structure `NintendoCreateAccountData` which would have to be added to the switch case. Though technically the function is already handling structures exclusive to friends like `AccountExtraInfo` and `NintendoLoginData`, since in other places I've only seen `AuthenticationInfo` being used (even in other NEX 1 games on the `RegisterEx` method)

We can add all of the above though since it may be useful for other purposes

DaniElectra:
> Hi! I'm working on my own NEX server for Fast Racing NEO ([MiniBioOne/fast-racing-neo](https://github.com/MiniBioOne/fast-racing-neo?rgh-link-date=2025-03-15T21%3A46%3A49Z)) and I'm confused on what exactly I need to add to my server to prepare for this PR. Just the environment variable (and code related to it)? And do I need to add any of those insecure login functions?

You can see how the friends server manages the AES key as a reference: https://github.com/PretendoNetwork/friends/blob/b44feb2ae0157f15569e94b6d6204f6e8c5c0779/init.go#L77

After that, you want to enable the Pretendo login of the method where the game sends the token, and enable the insecure method on the other protocol. On NEX 3 games for example you want to enable Pretendo validation on `TicketGranting` and enable the insecure `Register` method

globals/authentication.go
jonbarrow: You're right, I missed the part where the connection PID is validated so that would work fine there. So this could be used as-is if we decide to add support for `NintendoCreateAccountData`, which I think would be fine personally?

> Though technically the function is already handling structures exclusive to friends like AccountExtraInfo and NintendoLoginData, since in other places I've only seen AuthenticationInfo being used (even in other NEX 1 games on the RegisterEx method)

According to the wiki `AccountExtraInfo`, as a name, is not exclusive. It's used in the [account management protocol](https://nintendo-wiki.pretendo.network/docs/nex/protocols/account-management#accountextrainfo-structure) and [secure connection protocol](https://nintendo-wiki.pretendo.network/docs/nex/protocols/secure-connection/#nintendologindata--accountextrainfo)

Though it should be noted that even though the name is the same, the structure itself is different (this implementation asserts it to the account management version, but the secure connection version differs slightly)

DaniElectra: > According to the wiki AccountExtraInfo, as a name, is not exclusive. It's used in the [account management protocol](https://nintendo-wiki.pretendo.network/docs/nex/protocols/account-management#accountextrainfo-structure) and [secure connection protocol](https://nintendo-wiki.pretendo.network/docs/nex/protocols/secure-connection/#nintendologindata--accountextrainfo)
> 
> Though it should be noted that even though the name is the same, the structure itself is different (this implementation asserts it to the account management version, but the secure connection version differs slightly)

That is weird, I had tested the changes on the friends `RegisterEx` from 3DS and it was able to decode the token without issues. In fact, I believe the mention on `secure-connection` is wrong, since the decoding on the old nex-viewer shows some bytes of garbage at the beginning before the token string

jonbarrow: > That is weird, I had tested the changes on the friends `RegisterEx` from 3DS and it was able to decode the token without issues

I feel like that makes sense? It will decode just fine since that's handled at the protocol layer. All this function is doing is type asserting?

> In fact, I believe the mention on `secure-connection` is wrong, since the decoding on the old nex-viewer shows some bytes of garbage at the beginning before the token string

Would this not then mean that we've been decoding this structure wrong on our end then?

DaniElectra: > I feel like that makes sense? It will decode just fine since that's handled at the protocol layer. All this function is doing is type asserting?

What I mean is that on `SecureConnection::RegisterEx` it's using the `AccountManagement::AccountExtraInfo` structure, and not the "secure" one, so it has the fields of the `AccountManagement` one.

> Would this not then mean that we've been decoding this structure wrong on our end then?

On nex-protocols-go there is only the `AccountExtraInfo` from `AccountManagement`, so that part is already right. IDK about the new nex-viewer though

jonbarrow: > On nex-protocols-go there is only the `AccountExtraInfo` from `AccountManagement`, so that part is already right. IDK about the new nex-viewer though

What does the data look like in NEX Viewer? You said there was some garbage? If it lines up, I say that we just assume the wiki is wrong and remove the bad docs

MiniBioOne:
> > Hi! I'm working on my own NEX server for Fast Racing NEO ([MiniBioOne/fast-racing-neo](https://github.com/MiniBioOne/fast-racing-neo?rgh-link-date=2025-03-15T21%3A46%3A49Z)) and I'm confused on what exactly I need to add to my server to prepare for this PR. Just the environment variable (and code related to it)? And do I need to add any of those insecure login functions?
> 
> You can see how the friends server manages the AES key as a reference: https://github.com/PretendoNetwork/friends/blob/b44feb2ae0157f15569e94b6d6204f6e8c5c0779/init.go#L77
> 
> After that, you want to enable the Pretendo login of the method where the game sends the token, and enable the insecure method on the other protocol. On NEX 3 games for example you want to enable Pretendo validation on `TicketGranting` and enable the insecure `Register` method

~~I got the AES Key part and put it in globals, but how would I distinguish between the TicketGranting and Register method?~~

~~(it might just be that i'm being stupid and misunderstanding the PR)~~

Okay yeah I was just being dumb. I placed `commonSecureProtocol.EnableInsecureRegister()` in the register secure file and `commonTicketGrantingProtocol.SetPretendoValidation(globals.AESKey)` in the register auth file. Forgot that `Register` is a secure protocol method. Gonna test it soon.

MiniBioOne:
~~I apologise for the three commits, git got weird on asahi linux~~

EDIT: the commits from my repository appeared here, but I just recreated the repo as git screwed something up

globals/authentication.go
jonbarrow: Moving this back here from Discord: since we now know that when a new NEX account is created, the guest account uses the insecure `Register` method rather than `RegisterEx`, and the PIDs from the token and the connection will NOT match, it seems that the current function cannot be used as-is for the purposes of `NintendoCreateAccount` since:

- We cannot take the PID of the connection as that will be the guest account (PID 100), meaning the function WOULD need to return the token data
- Since the function compares the connection PID to the token PID, guest account logins would error here anyway

We have 2 real options then:

1. Duplicate all the logic here in `NintendoCreateAccount`, but remove the PID check (would reduce code duplication, at the cost of the issues mentioned below)
2. Update the function so that the guest account PID is also valid, and return the token data (would reduce code duplication, but could introduce errors/bugs that may allow for the guest account more access than it should have, and would bloat the function)

Thoughts?

DaniElectra: I think we could do 1 here and leave guest account access checks for later when we redesign the protocol handlers in nex-go as mentioned on Discord, where we could add our own checks before method handlers get called

jonbarrow: That sounds fine to me 👍

DaniElectra:
This PR has been approved on Discord

PR #54, opened 2025-06-14
feat: Implement MessageDelivery and Messaging protocols

DaniElectra:
Resolves #XXX

### Changes:

This PR aims to implement complete common protocols of Messaging and MessageDelivery. At the moment, only  MessageDelivery::DeliverMessage has been added as a base, along with instant messages. This is done to allow reviewing the design and structure that the common protocols will have. **These changes haven't been tested, and have only been pushed for allowing review of the current design.**

### Task list

- `Messaging`
  - [x] `Messaging::DeliverMessage`
  - [x] `Messaging::GetNumberOfMessages`
  - [x] `Messaging::GetMessagesHeaders`
  - [x] `Messaging::RetrieveAllMessagesWithinRange`
  - [x] `Messaging::RetrieveMessages`
  - [x] `Messaging::DeleteMessages`
  - [x] `Messaging::DeleteAllMessages`
  - [x] `Messaging::DeliverMessageMultiTarget`

- `MessageDelivery`
  - [x] `MessageDelivery::DeliverMessage`
  - [x] `MessageDelivery::DeliverMessageMultiTarget`

### Design

Following the design of common Matchmaking, these protocols will have a shared `MessagingManager` that the protocols will be able to access in order to interact with the underlying database that will store the messages. This manager has been designed with modularity in mind, allowing customizations like extensions to recipient types or additional structures that inherit from `UserMessage` to integrate on the existing code with little complexity.

In order to achieve this goal, the `MessagingManager` allows changing certain functions used on different contexts. The following is a list of those functions, and may be expanded as more methods and functionality get implemented:
- `ValidateMessage`: Used for validating the fields of a message, allowing integration of custom message types
- `GetMessageHeader`: Used for retrieving the common `UserMessage` header from a specific message type
- `SetMessageHeader`: Used for setting the common `UserMessage` header on a specific message type
- `ProcessMessage`: Called when delivering a new message

For each of those functions a default implementation is given by either the `MessagingManager` itself or the common protocols, in order to accomodate the necessity of accessing functionality of other common protocols like matchmaking. Following that, the `MessagingManager` includes an **optional** parameter for a `MatchmakingManager`, which will be used on games that send messages to a gathering.

The database functionality is implemented under the `messaging` protocol, while `message-delivery` will access it using the same functions as `messaging`, in a similar fashion to how `match-making-ext` interacts with matchmaking functionality using the functions implemented under `match-making`. The database creation functions will probably be duplicated in both protocols in order to handle games which will only use one of the protocols, like exclusively using `MessageDelivery`, for example.

Finally, in order to handle changes made to the `UserMessage` structure on NEX 4.0 and later where the message recipient information was moved from individual fields to a `MessageRecipient` parameter, a getter `GetUserMessageRecipientData` and setter is implemented for accessing this information while handling versioning differences.

Regarding the database, it will have the following structure:
- `messaging` as the main schema. Inside it there will be tables for handling various functionality, like:
  - `messages` for the common base `UserMessage` of *non-instant* messages. Instant messages have a separate table since those technically don't have an ID assigned, but we will store them for tracking purposes
  - `text_messages` for storing the `TextMessage`-specific data, and will be linked to the `messages` table by ID
  - `binary_messages` for storing the `BinaryMessage`-specific data, and will be linked to the `messages` table by ID
  - `instant_messages` for the common base `UserMessage` of *instant* messages
  - `instant_text_messages` for storing the `TextMessage`-specific data, and will be linked to the `instant_messages` table by ID
  - `instant_binary_messages` for storing the `BinaryMessage`-specific data, and will be linked to the `instant_messages` table by ID

If you want to test these changes, you can use code similar to the following, adjusting it for your necessities:

```go
messagingManager := common_globals.NewMessagingManager(globals.SecureEndpoint, database.Postgres)

messageDeliveryProtocol := message_delivery.NewProtocol()
globals.SecureEndpoint.RegisterServiceProtocol(messageDeliveryProtocol)
commonMessageDeliveryProtocol := common_message_melivery.NewCommonProtocol(messageDeliveryProtocol)
commonMessageDeliveryProtocol.SetManager(messageDeliveryManager)
```

### DIFF
--- a/message-delivery/deliver_message_multi_target.go
+++ b/message-delivery/deliver_message_multi_target.go
@@ -8,7 +8,7 @@ import (
 	message_delivery "github.com/PretendoNetwork/nex-protocols-go/v2/message-delivery"
 )
 
-func (commonProtocol *CommonProtocol) deliverMessage(err error, packet nex.PacketInterface, callID uint32, oUserMessage types.DataHolder) (*nex.RMCMessage, *nex.Error) {
+func (commonProtocol *CommonProtocol) deliverMessageMultiTarget(err error, packet nex.PacketInterface, callID uint32, lstTarget types.List[types.PID], oUserMessage types.DataHolder) (*nex.RMCMessage, *nex.Error) {
 	if err != nil {
 		common_globals.Logger.Error(err.Error())
 		return nil, nex.NewError(nex.ResultCodes.Core.InvalidArgument, err.Error())
@@ -20,10 +20,10 @@ func (commonProtocol *CommonProtocol) deliverMessage(err error, packet nex.Packe
 	// NOTE - This method will silently fail on any errors
 	rmcResponse := nex.NewRMCSuccess(endpoint, nil)
 	rmcResponse.ProtocolID = message_delivery.ProtocolID
-	rmcResponse.MethodID = message_delivery.MethodDeliverMessage
+	rmcResponse.MethodID = message_delivery.MethodDeliverMessageMultiTarget
 	rmcResponse.CallID = callID
 
-	recipientID, recipientType, nexError := commonProtocol.manager.ValidateMessage(oUserMessage)
+	_, _, nexError := commonProtocol.manager.ValidateMessage(oUserMessage)
 	if nexError != nil {
 		common_globals.Logger.Error(nexError.Error())
 		return rmcResponse, nil
@@ -35,7 +35,12 @@ func (commonProtocol *CommonProtocol) deliverMessage(err error, packet nex.Packe
 		return rmcResponse, nil
 	}
 
-	_, _, _, nexError = commonProtocol.manager.ProcessMessage(commonProtocol.manager, oUserMessage, types.List[types.UInt64]{recipientID}, recipientType, true)
+	recipientIDs := make(types.List[types.UInt64], len(lstTarget))
+	for i, recipientID := range lstTarget {
+		recipientIDs[i] = types.UInt64(recipientID)
+	}
+
+	_, _, _, nexError = commonProtocol.manager.ProcessMessage(commonProtocol.manager, oUserMessage, recipientIDs, 1, true)
 	if nexError != nil {
 		common_globals.Logger.Error(nexError.Error())
 		return rmcResponse, nil
--- a/message-delivery/deliver_message_multi_target.go
+++ b/message-delivery/deliver_message_multi_target.go
@@ -23,6 +23,12 @@ func (commonProtocol *CommonProtocol) deliverMessageMultiTarget(err error, packe
 	rmcResponse.MethodID = message_delivery.MethodDeliverMessageMultiTarget
 	rmcResponse.CallID = callID
 
+	// * Only allow up to 100 targets, based on the maximum amount of friends allowed
+	if len(lstTarget) > 100 {
+		common_globals.Logger.Error("Message has over 100 targets")
+		return rmcResponse, nil
+	}
+
 	_, _, nexError := commonProtocol.manager.ValidateMessage(oUserMessage)
 	if nexError != nil {
 		common_globals.Logger.Error(nexError.Error())
--- a/message-delivery/protocol.go
+++ b/message-delivery/protocol.go
@@ -1,0 +1,83 @@
+package message_delivery
+
+import (
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	message_delivery "github.com/PretendoNetwork/nex-protocols-go/v2/message-delivery"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+	messaging_database "github.com/PretendoNetwork/nex-protocols-common-go/v2/messaging/database"
+)
+
+type CommonProtocol struct {
+	endpoint              nex.EndpointInterface
+	protocol              message_delivery.Interface
+	manager               *common_globals.MessagingManager
+	OnAfterDeliverMessage func(packet nex.PacketInterface, oUserMessage types.DataHolder)
+}
+
+// SetManager defines the messaging manager to be used by the common protocol
+func (commonProtocol *CommonProtocol) SetManager(manager *common_globals.MessagingManager) {
+	var err error
+
+	commonProtocol.manager = manager
+
+	if manager.ProcessMessage == nil {
+		manager.ProcessMessage = func(message types.DataHolder, recipientID types.UInt64, recipientType types.UInt32, sendMessage bool) (types.DataHolder, types.List[types.UInt32], types.List[types.PID], *nex.Error) {
+			return messaging_database.ProcessMessage(manager, message, recipientID, recipientType, sendMessage)
+		}
+	}
+
+	_, err = manager.Database.Exec(`CREATE SCHEMA IF NOT EXISTS messaging`)
+	if err != nil {
+		common_globals.Logger.Error(err.Error())
+		return
+	}
+
+	_, err = manager.Database.Exec(`CREATE TABLE IF NOT EXISTS messaging.instant_messages (
+		id bigserial PRIMARY KEY,
+		recipient_id numeric(20),
+		recipient_type numeric(10),
+		parent_id bigint,
+		sender_pid numeric(20),
+		reception_time timestamp,
+		lifetime numeric(10),
+		flags bigint,
+		subject text,
+		sender text,
+		type text NOT NULL DEFAULT ''
+	)`)
+	if err != nil {
+		common_globals.Logger.Error(err.Error())
+		return
+	}
+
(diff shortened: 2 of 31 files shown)


TraceEntertains:
Have tested this implementation of MessageDelivery (namely, DeliverMessage) in a couple games (yo-kai watch 3 and yo-kai watch blasters iirc?) and seems to work quite well

DaniElectra:
@TraceEntertains has tested this PR and confirmed it works. Opening for review now

message-delivery/deliver_message_multi_target.go
```diff
@@ -0,0 +1,50 @@
...
+	oUserMessage, nexError = commonProtocol.manager.PrepareMessage(connection.PID(), oUserMessage)
+	if nexError != nil {
+		common_globals.Logger.Error(nexError.Error())
+		return rmcResponse, nil
+	}
+
+	recipientIDs := make(types.List[types.UInt64], len(lstTarget))
+	for i, recipientID := range lstTarget {
+		recipientIDs[i] = types.UInt64(recipientID)
+	}
```
jonbarrow, line 47: Should we cap this? Also why do we have to remake the list, `types.PID` is an alias of `uint64` same as `types.UInt64`?

message-delivery/protocol.go
```diff
@@ -0,0 +1,125 @@
...
+	_, err = manager.Database.Exec(`CREATE TABLE IF NOT EXISTS messaging.instant_binary_messages (
+		id bigint PRIMARY KEY REFERENCES messaging.instant_messages(id),
+		body bytea
+	)`)
+	if err != nil {
+		common_globals.Logger.Error(err.Error())
+		return
+	}
+
+	_, err = manager.Database.Exec(`CREATE TABLE IF NOT EXISTS messaging.messages (
```
jonbarrow, line 76: To line up with `instant_messages`, maybe this should be named `persistent_messages`?

messaging/database/delete_all_messages.go
```diff
@@ -0,0 +1,31 @@
...
+func DeleteAllMessages(manager *common_globals.MessagingManager, recipientID types.UInt64, recipientType types.UInt32) (types.UInt32, *nex.Error) {
+	var err error
+
+	result, err := manager.Database.Exec(`UPDATE FROM messaging.messages SET deleted = true WHERE
+		recipient_id = $1 AND
+		recipient_type = $2 AND
+		reception_time + lifetime * INTERVAL '1 second' > NOW() AND
+		read IS NOT TRUE AND
+		deleted IS NOT TRUE
+	`, recipientID, recipientType)
```
jonbarrow, line 20: This query won't delete a message if it's already been read or if it's expired, is that intended? That seems odd that you can't delete read/expired messages?

messaging/database/delete_messages.go
```diff
@@ -0,0 +1,27 @@
...
+func DeleteMessages(manager *common_globals.MessagingManager, recipientID types.UInt64, recipientType types.UInt32, lstMessagesToDelete types.List[types.UInt32]) *nex.Error {
+	var err error
+
+	_, err = manager.Database.Exec(`UPDATE FROM messaging.messages SET deleted = true WHERE
+		id = ANY($1) AND
+		recipient_id = $2 AND
+		recipient_type = $3 AND
+		reception_time + lifetime * INTERVAL '1 second' > NOW() AND
+		read IS NOT TRUE AND
+		deleted IS NOT TRUE
```
jonbarrow, line 20: Same question as from `DeleteAllMessages`

messaging/database/insert_binary_message.go
```diff
@@ -0,0 +1,31 @@
...
+	) VALUES (
+		$1,
+		$2
+	)`, messageID, message.BinaryBody)
+	if err != nil {
+		return nex.NewError(nex.ResultCodes.Core.Unknown, err.Error())
+	}
+
+	return nil
+}
```
jonbarrow, line 59: I've been thinking about how we handle certain database actions to prevent load. Currently this function alone triggers 2 queries, which is 2 connections to the database. If we use `WITH ... AS` we can do this in a single query by using a subquery that does the 1st insert and returns the ID, for use in the main query

Maybe that's worth exploring here since these queries seem relatively simple? See how it works for us?

messaging/database/insert_text_message.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	messaging_types "github.com/PretendoNetwork/nex-protocols-go/v2/messaging/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+)
+
+// InsertTextMessage inserts a new text message into the database
+func InsertTextMessage(manager *common_globals.MessagingManager, message messaging_types.TextMessage, recipientID types.UInt64, recipientType types.UInt32) *nex.Error {
+	var err error
+
+	messageID, nexError := InsertInstantMessage(manager, message.UserMessage, recipientID, recipientType, "TextMessage")
```
jonbarrow: Should this be `InsertMessage`?

messaging/database/delete_all_messages.go
DaniElectra: When I was testing the "delete messages" method it seems that it would actually remove the entry from the database, since it's ID could be assigned to a new message.

However, nothing would happen when trying to delete a read or expired message, so here I'm sort of emulating that behavior even though we won't actually delete messages in any case

messaging/database/insert_text_message.go
DaniElectra: That's correct, good catch

message-delivery/deliver_message_multi_target.go
DaniElectra: > Should we cap this?

Probably yeah. We don't have a way to know what original servers capped this to (or if they even capped it in the first place), so I'm open to suggestions (maybe up to 100 matching the friends limit?)

> Also why do we have to remake the list, types.PID is an alias of uint64 same as types.UInt64?

Those are type *definitions* (not type *aliases*), and we can't convert lists of different type definitions even if they originate from the same type, so we gotta do this

messaging/database/insert_binary_message.go
DaniElectra: That can be consiedered yeah

message-delivery/deliver_message_multi_target.go
jonbarrow: Adding a cap of 100 sounds fine

As for the types, you can't directly cast them but you can add a type constraint to the function to allow any type whose underlying type is a given value. For example, this works:

```go
package main

import "fmt"

type List[T any] []T
type UInt64 uint64
type PID uint64

func main() {
	list := make(List[PID], 0)
	listConsumer(list)
}

func listConsumer[T ~uint64](list List[T]) {
	fmt.Println(list)
}
```

message-delivery/protocol.go
DaniElectra: I think `messages` already implies they are stored in a "persistent" way, especially since we have a different table for "instant" messages

I'm not against this though, I can rename it if really needed

message-delivery/deliver_message_multi_target.go
DaniElectra: That doesn't seem to work because `ProcessMessage` is defined as an overridable function on `MessagingManager`, and I can't seem to use generics on the function inside the struct

jonbarrow: You're right I forgot about it being weird on structs

It can still work if you define the generic on the struct itself, but that's kinda ugly for a one-off function

Another solution is to make an intermediate type like

```go
type SomeMiddleTypeIdkTheName interface {
	~uint64
}
```

And use THAT as the list type accepted by the method. But again, might be a bit ugly for a one-off

DaniElectra: That doesn't work either as it tells me I can't use the interface outside a type constraint (generics)

jonbarrow: You're right, I was mistaken. That only works if the `SomeMiddleTypeIdkTheName` is used as a generic on the function still, which doesn't solve the problem either

Either way, I don't think it's worth the hassle right now. I was trying to save on list creations, but it's not that big of deal. Especially if we just cap how big the input list is so it doesn't take forever

message-delivery/protocol.go
jonbarrow: On second thought, it's fine the way it is I agree

messaging/database/delete_all_messages.go
jonbarrow: Fair enough 👍

messaging/database/insert_binary_message.go
jonbarrow: I won't block the PR over it, but it would be welcome if you wanted to take a look at it 👍

jonbarrow (approved):
LGTM

PR #55, opened 2025-06-14
feat(matchmake-extension): Implement GetPlayingSession

DaniElectra:
Resolves #XXX

### Changes:

Reuse the code from the `GetSimplePlayingSession` method, assuming it behaves in a similar way. Also make the cleanup functions optional, as they are only intended for special cases like region locking in a game or similar.

Haven't tested the changes tbh, but I trust myself that this should be a simple change and work as expected.


TraceEntertains:
IIRC this has been tested in a couple games now, and seems to work fine, there is a strange bug in one game tested, but it is unlikely to have anything to do with this method lol (game doesn't show gatherings in a list even when server returns one), but in the same game with different MatchmakeSession info it works perfectly fine, output has been confirmed to be 1 to 1 for the original (minus structure versions)

matchmake-extension/get_playing_session.go
```diff
@@ -0,0 +1,51 @@
...
+		common_globals.Logger.Error(err.Error())
+		return nil, nex.NewError(nex.ResultCodes.Core.InvalidArgument, err.Error())
+	}
+
+	connection := packet.Sender().(*nex.PRUDPConnection)
+	endpoint := connection.Endpoint().(*nex.PRUDPEndPoint)
+
+	commonProtocol.manager.Mutex.RLock()
+
+	playingSessions, nexError := database.GetPlayingSession(commonProtocol.manager, lstPID)
```
jonbarrow: `lstPID` is missing a length check. Only up to [`MAX_PRINCIPALID_SIZE_TO_FIND_MATCHMAKE_SESSION`](https://github.com/tech-ticks/RTDXTools/blob/232a7797e01369e9c12704f58fdde65dd3ac1c32/Assets/Scripts/Stubs/Generated/Assembly-CSharp/NexPlugin/Matchmake.cs#L30) can be requested at a time. I forgot which error is returned but `InvalidArgument` should be fine

matchmake-extension/database/get_playing_session.go
```diff
@@ -0,0 +1,93 @@
...
+
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+	match_making_types "github.com/PretendoNetwork/nex-protocols-go/v2/match-making/types"
+)
+
+// GetPlayingSession returns the playing sessions of the given PIDs
+func GetPlayingSession(manager *common_globals.MatchmakingManager, listPID []types.PID) ([]match_making_types.PlayingSession, *nex.Error) {
+	playingSessions := make([]match_making_types.PlayingSession, 0)
```
jonbarrow: In my testing only up to [`MAX_MATCHMAKE_SESSION_BY_PARTICIPANT`](https://github.com/tech-ticks/RTDXTools/blob/232a7797e01369e9c12704f58fdde65dd3ac1c32/Assets/Scripts/Stubs/Generated/Assembly-CSharp/NexPlugin/Matchmake.cs#L32) is returned

matchmake-extension/database/get_playing_session.go
```diff
@@ -0,0 +1,93 @@
...
+			ms.refer_gid,
+			ms.user_password_enabled,
+			ms.system_password_enabled,
+			ms.codeword
+		FROM matchmaking.gatherings AS g
+		INNER JOIN matchmaking.matchmake_sessions AS ms ON ms.id = g.id
+		WHERE
+		g.registered=true AND
+		g.type='MatchmakeSession' AND
+		$1=ANY(g.participants)`, pid).Scan(
```
jonbarrow: You can get the session info for anyone, but if you request info for a user who has you blocked then that users info is not returned. At the very least we should make a note of this somewhere, if not adding it now

matchmake-extension/get_playing_session.go
```diff
@@ -0,0 +1,51 @@
...
+	if nexError != nil {
+		commonProtocol.manager.Mutex.RUnlock()
+		return nil, nexError
+	}
+
+	commonProtocol.manager.Mutex.RUnlock()
+
+	lstPlayingSession := types.NewList[match_making_types.PlayingSession]()
+	lstPlayingSession = playingSessions
```
jonbarrow: You can avoid this by just returning the `types.List` from the database method. The `DataStore` rework PR does this in a few places. `types.List` can be used as a normal slice/array type

jonbarrow (changes requested):
Looks good! Just missing a few sanity checks/notes

jonbarrow:
LGTM! I'll approve this now but won't merge yet until given the greenlight, incase someone wants to test the latest changes

jonbarrow (dismissed):

matchmake-extension/database/get_playing_session.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+	match_making_types "github.com/PretendoNetwork/nex-protocols-go/v2/match-making/types"
+)
+
+// GetPlayingSession returns the playing sessions of the given PIDs
+func GetPlayingSession(manager *common_globals.MatchmakingManager, listPID types.List[types.PID]) (types.List[match_making_types.PlayingSession], *nex.Error) {
+	playingSessions := make([]match_making_types.PlayingSession, 0)
```
jonbarrow: I didn't catch this actually, this can be improved. In Go if a slice is made without any capacity set then whenever you `append` to the slice it will double the capacity, reallocate the memory, and then copy the contents of the old slice to the new one. Since we know that only 1000 entries can exist here, we can go ahead and pre-allocate the slice to avoid this overhead. Preallocating the slice *could* waste some memory for slices smaller than 1000, but it avoids the need for multiple reallocations and copies

DaniElectra: Ah right, I didn't add it at first since it would be wasting memory, but I'll go ahead and add it

TraceEntertains:
Tested on Yo-kai Watch Blasters (before slice pre-allocation change, but I don't think that would break it), seems to work fine

jonbarrow (approved):

PR #56, opened 2025-07-09
feat: NEX 4.x.x compatibility

TraceEntertains:
Resolves #XXX

### Changes:

Fix TicketGranting, UpdateNotificationData methods for NEX 4.x.x, Fix NotificationData structure for NEX 4.x.x.
~~fix some attribute 1 errors in FindMatchmakeSessionBySearchCriteria (yes this last one isn't NEX 4 but it is a very useful fix)~~ Removed from this pull request, will create new one later

- [X] I have tested all of my changes.

### DIFF
--- a/match-making/database/disconnect_participant.go
+++ b/match-making/database/disconnect_participant.go
@@ -49,7 +49,7 @@ func DisconnectParticipant(manager *common_globals.MatchmakingManager, connectio
 		// * If the gathering is a PersistentGathering and the gathering isn't set to leave when disconnecting, ignore and continue
 		//
 		// TODO - Is the match_making.GatheringFlags.PersistentGathering check correct here?
-		if uint32(gathering.Flags) & match_making.GatheringFlags.PersistentGathering != 0 || (gatheringType == "PersistentGathering" && uint32(gathering.Flags) & match_making.GatheringFlags.PersistentGatheringLeaveParticipation == 0) {
+		if uint32(gathering.Flags)&match_making.GatheringFlags.PersistentGathering != 0 || (gatheringType == "PersistentGathering" && uint32(gathering.Flags)&match_making.GatheringFlags.PersistentGatheringLeaveParticipation == 0) {
 			continue
 		}
 
@@ -67,7 +67,7 @@ func DisconnectParticipant(manager *common_globals.MatchmakingManager, connectio
 		}
 
 		// * If the gathering is a persistent gathering and allows zero users, only remove the participant from the gathering
-		if uint32(gathering.Flags) & (match_making.GatheringFlags.PersistentGathering | match_making.GatheringFlags.PersistentGatheringAllowZeroUsers) != 0 {
+		if uint32(gathering.Flags)&(match_making.GatheringFlags.PersistentGathering|match_making.GatheringFlags.PersistentGatheringAllowZeroUsers) != 0 {
 			continue
 		}
 
@@ -87,7 +87,7 @@ func DisconnectParticipant(manager *common_globals.MatchmakingManager, connectio
 			// * This flag tells the server to change the matchmake session owner if they disconnect
 			// * If the flag is not set, delete the session
 			// * More info: https://nintendo-wiki.pretendo.network/docs/nex/protocols/match-making/types#flags
-			if uint32(gathering.Flags) & match_making.GatheringFlags.DisconnectChangeOwner == 0 {
+			if uint32(gathering.Flags)&match_making.GatheringFlags.DisconnectChangeOwner == 0 {
 				nexError = UnregisterGathering(manager, connection.PID(), uint32(gathering.ID))
 				if nexError != nil {
 					common_globals.Logger.Error(nexError.Error())
@@ -99,8 +99,8 @@ func DisconnectParticipant(manager *common_globals.MatchmakingManager, connectio
 
 				oEvent := notifications_types.NewNotificationEvent()
 				oEvent.PIDSource = connection.PID().Copy().(types.PID)
-				oEvent.Type = types.NewUInt32(notifications.BuildNotificationType(category, subtype))
-				oEvent.Param1 = gathering.ID
+				oEvent.Type = types.UInt32(notifications.BuildNotificationType(category, subtype))
+				oEvent.Param1 = types.UInt64(gathering.ID)
 
 				common_globals.SendNotificationEvent(connection.Endpoint().(*nex.PRUDPEndPoint), oEvent, common_globals.RemoveDuplicates(participants))
 
@@ -125,8 +125,8 @@ func DisconnectParticipant(manager *common_globals.MatchmakingManager, connectio
 
 				oEvent := notifications_types.NewNotificationEvent()
 				oEvent.PIDSource = connection.PID().Copy().(types.PID)
-				oEvent.Type = types.NewUInt32(notifications.BuildNotificationType(category, subtype))
-				oEvent.Param1 = gathering.ID
+				oEvent.Type = types.UInt32(notifications.BuildNotificationType(category, subtype))
+				oEvent.Param1 = types.UInt64(gathering.ID)
 
 				// TODO - Should the notification actually be sent to all participants?
 				common_globals.SendNotificationEvent(connection.Endpoint().(*nex.PRUDPEndPoint), oEvent, common_globals.RemoveDuplicates(participants))
@@ -144,14 +144,14 @@ func DisconnectParticipant(manager *common_globals.MatchmakingManager, connectio
 
 		oEvent := notifications_types.NewNotificationEvent()
 		oEvent.PIDSource = connection.PID().Copy().(types.PID)
-		oEvent.Type = types.NewUInt32(notifications.BuildNotificationType(category, subtype))
-		oEvent.Param1 = gathering.ID
-		oEvent.Param2 = types.NewUInt32(uint32(connection.PID())) // TODO - This assumes a legacy client. Will not work on the Switch
+		oEvent.Type = types.UInt32(notifications.BuildNotificationType(category, subtype))
+		oEvent.Param1 = types.UInt64(gathering.ID)
+		oEvent.Param2 = types.UInt64(connection.PID())
 
 		var participantEndedTargets []uint64
 
 		// * When the VerboseParticipants or VerboseParticipantsEx flags are set, all participant notification events are sent to everyone
-		if uint32(gathering.Flags) & (match_making.GatheringFlags.VerboseParticipants | match_making.GatheringFlags.VerboseParticipantsEx) != 0 {
+		if uint32(gathering.Flags)&(match_making.GatheringFlags.VerboseParticipants|match_making.GatheringFlags.VerboseParticipantsEx) != 0 {
 			participantEndedTargets = common_globals.RemoveDuplicates(participants)
 		} else {
 			participantEndedTargets = []uint64{uint64(gathering.OwnerPID)}
--- a/matchmake-extension/database/update_notification_data.go
+++ b/matchmake-extension/database/update_notification_data.go
@@ -2,8 +2,8 @@ package database
 
 import (
 	"github.com/PretendoNetwork/nex-go/v2"
-	notifications_types "github.com/PretendoNetwork/nex-protocols-go/v2/notifications/types"
 	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+	notifications_types "github.com/PretendoNetwork/nex-protocols-go/v2/notifications/types"
 )
 
 // UpdateNotificationData updates the notification data of the specified user and type
--- a/matchmake-extension/protocol.go
+++ b/matchmake-extension/protocol.go
@@ -11,37 +11,37 @@ import (
 )
 
 type CommonProtocol struct {
-	endpoint                                         nex.EndpointInterface
-	protocol                                         matchmake_extension.Interface
-	manager                                          *common_globals.MatchmakingManager
-	PersistentGatheringCreationMax                   int
-	CanJoinMatchmakeSession                          func(manager *common_globals.MatchmakingManager, pid types.PID, matchmakeSession match_making_types.MatchmakeSession) *nex.Error
-	CleanupSearchMatchmakeSession                    func(matchmakeSession *match_making_types.MatchmakeSession)
-	CleanupMatchmakeSessionSearchCriterias           func(searchCriterias types.List[match_making_types.MatchmakeSessionSearchCriteria])
-	OnAfterOpenParticipation                         func(packet nex.PacketInterface, gid types.UInt32)
-	OnAfterCloseParticipation                        func(packet nex.PacketInterface, gid types.UInt32)
-	OnAfterCreateMatchmakeSession                    func(packet nex.PacketInterface, anyGathering match_making_types.GatheringHolder, message types.String, participationCount types.UInt16)
-	OnAfterGetSimplePlayingSession                   func(packet nex.PacketInterface, listPID types.List[types.PID], includeLoginUser types.Bool)
-	OnAfterAutoMatchmakePostpone                     func(packet nex.PacketInterface, anyGathering match_making_types.GatheringHolder, message types.String)
-	OnAfterAutoMatchmakeWithParamPostpone            func(packet nex.PacketInterface, autoMatchmakeParam match_making_types.AutoMatchmakeParam)
-	OnAfterAutoMatchmakeWithSearchCriteriaPostpone   func(packet nex.PacketInterface, lstSearchCriteria types.List[match_making_types.MatchmakeSessionSearchCriteria], anyGathering match_making_types.GatheringHolder, strMessage types.String)
-	OnAfterGetPlayingSession                         func(packet nex.PacketInterface, lstPID types.List[types.PID])
-	OnAfterCreateCommunity                           func(packet nex.PacketInterface, community match_making_types.PersistentGathering, strMessage types.String)
-	OnAfterFindCommunityByGatheringID                func(packet nex.PacketInterface, lstGID types.List[types.UInt32])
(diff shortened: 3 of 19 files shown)


jonbarrow:
I'll give this a proper review in a bit, but in the future can you please make smaller, more focused, commits that each change specific things, rather than single commits that touch many areas, when possible? Same thing with the PRs themselves, the NEX 4 changes should be in a separate PR from the general matchmaking changes so that we are trying to review multiple "types" of changes all at once

Since there's only a single commit right now I THINK this can still be fixed without much issue, by reverting the changes, making new commits, and force-pushing back? I'll CC @DaniElectra for thoughts on that. If not it's not a huge deal, just something for the future then

TraceEntertains:
Removed the attribute 1 fix from this, sorry about not splitting the notification stuff and ticketgranting stuff into separate commits, I kinda forgot and don't think it warrants a second force push 😭

matchmake-extension/database/update_notification_data.go
```diff
@@ -2,8 +2,8 @@ package database
 
 import (
 	"github.com/PretendoNetwork/nex-go/v2"
-	notifications_types "github.com/PretendoNetwork/nex-protocols-go/v2/notifications/types"
 	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+	notifications_types "github.com/PretendoNetwork/nex-protocols-go/v2/notifications/types"
```
DaniElectra, line 6: This is a useless change, can you undo it?

match-making/database/disconnect_participant.go
```diff
@@ -49,7 +49,7 @@ func DisconnectParticipant(manager *common_globals.MatchmakingManager, connectio
 		// * If the gathering is a PersistentGathering and the gathering isn't set to leave when disconnecting, ignore and continue
 		//
 		// TODO - Is the match_making.GatheringFlags.PersistentGathering check correct here?
-		if uint32(gathering.Flags) & match_making.GatheringFlags.PersistentGathering != 0 || (gatheringType == "PersistentGathering" && uint32(gathering.Flags) & match_making.GatheringFlags.PersistentGatheringLeaveParticipation == 0) {
+		if uint32(gathering.Flags)&match_making.GatheringFlags.PersistentGathering != 0 || (gatheringType == "PersistentGathering" && uint32(gathering.Flags)&match_making.GatheringFlags.PersistentGatheringLeaveParticipation == 0) {
```
DaniElectra, line 52: Why are we changing all of these?

matchmake-extension/protocol.go
```diff
@@ -107,8 +107,8 @@ func (commonProtocol *CommonProtocol) SetManager(manager *common_globals.Matchma
 		id bigserial PRIMARY KEY,
 		source_pid numeric(10),
 		type bigint,
-		param_1 bigint,
-		param_2 bigint,
+		param_1 numeric(10),
```
DaniElectra: This is technically incorrect, `numeric(10)` doesn't cover all uint64 values, that would be `numeric(20)`. That is a mistake I made when I was working on this code, and it has remained since then 😅

DaniElectra (changes requested):
Important note that the database update is a breaking change. It won't affect current databases (since games we currently support only use uint32 values), but we should consider doing something about it

match-making/database/disconnect_participant.go
TraceEntertains: That was ***entirely*** unintentional, my vscode keeps doing it and I don't know how to turn it off 😭

matchmake-extension/database/update_notification_data.go
TraceEntertains: Oops, didn't catch that one, will remove it

matchmake-extension/protocol.go
TraceEntertains: lol 😭

TraceEntertains: Is it fine if I just... swap out all the numeric(10)s with numeric(20)s then?

TraceEntertains:
> Important note that the database update is a breaking change. It won't affect current databases (since games we currently support only use uint32 values), but we should consider doing something about it

I have literally no ideas for how to avoid breaking databases so I am open to ideas :P

matchmake-extension/database/update_notification_data.go
TraceEntertains: Actually, my vscode also just keeps doing that

matchmake-extension/protocol.go
jonbarrow: > Is it fine if I just... swap out all the numeric(10)s with numeric(20)s then?

If the database already exists then changing the column type alone won't do anything at all. A migration would need to also happen

match-making/database/disconnect_participant.go
jonbarrow: If you're getting changes on save then it's whatever formatter you have. Open your VSC settings and search for "go format" and see what you have set for the "Format Tool" setting, and also check if the language server is enabled. Whatever formatter you're using is causing this

ticket-granting/login.go
```diff
@@ -61,7 +62,7 @@ func (commonProtocol *CommonProtocol) login(err error, packet nex.PacketInterfac
 		pConnectionData.StationURLSpecialProtocols = commonProtocol.StationURLSpecialProtocols
 		pConnectionData.Time = types.NewDateTime(0).Now()
 
-		if endpoint.LibraryVersions().Main.GreaterOrEqual("v3.5.0") {
+		if server.LibraryVersions.Main.GreaterOrEqual("3.5.0") {
```
jonbarrow, line 69: Why was this changed? A server can have multiple PRUDP endpoints and each endpoint can be configured with it's own settings. A packet should be using the settings for the endpoint it is assigned to

ticket-granting/login_ex.go
```diff
@@ -75,6 +79,10 @@ func (commonProtocol *CommonProtocol) loginEx(err error, packet nex.PacketInterf
 	pConnectionData.WriteTo(rmcResponseStream)
 	strReturnMsg.WriteTo(rmcResponseStream)
 
+	if server.LibraryVersions.Main.GreaterOrEqual("4.0.0") {
+		pSourceKey.WriteTo(rmcResponseStream)
+	}
```
jonbarrow, line 92: Is this accurate? The wiki states that this is only present on the Switch, not that it's a NEX version difference

CC @DaniElectra

ticket-granting/login_ex.go
```diff
@@ -44,6 +45,9 @@ func (commonProtocol *CommonProtocol) loginEx(err error, packet nex.PacketInterf
 	pConnectionData := types.NewRVConnectionData()
 	strReturnMsg := types.NewString("")
 
+	// TODO - Does pSourceKey need to be set for anything?
```
jonbarrow: `pSourceKey` is used to skip the key derivation for decrypting the ticket. If this is a non-empty hex string, then it is used as the key instead. It doesn't HAVE to be set to anything if we never intend for this to be supported, but it would be nice to at least have it configurable

ticket-granting/login_ex.go
```diff
@@ -44,6 +45,9 @@ func (commonProtocol *CommonProtocol) loginEx(err error, packet nex.PacketInterf
...
+	pSourceKey := types.NewString("")
```
jonbarrow, line 54: Mostly a style nit, but you defined this parameter away from the other response parameters. We tend to prefer that these get grouped together, and that any comments just get put next to them to maintain that grouping

jonbarrow (changes requested):
Looks good so far, thank you! Just a couple minor things outside of what @DaniElectra brought u[

ticket-granting/login_ex.go
TraceEntertains: I confirmed in dev-meta with Dani that this is a NEX version difference, and not a switch-only thing (the NEX 4.2.0 3ds game that was used to test would not go past TicketGranting without pSourceKey in the response)

The wiki probably does need to be updated

ticket-granting/login_ex.go
TraceEntertains: Yeah, sure, I can change that

ticket-granting/login.go
TraceEntertains: Dani said that for consistency, I should use the implementation that [JoinMatchmakeSession](https://github.com/PretendoNetwork/nex-protocols-common-go/blob/50586821d228cdc2bae728797febdae51e67fe5b/matchmake-extension/join_matchmake_session.go#L62) uses, so I changed it for Login as well due to it being in TicketGranting next to LoginEx and RequestTicket

matchmake-extension/protocol.go
TraceEntertains: > > Is it fine if I just... swap out all the numeric(10)s with numeric(20)s then?
> 
> If the database already exists then changing the column type alone won't do anything at all. A migration would need to also happen

I'm well aware 😭

I don't really know if we would be handling that in code or manually though

match-making/database/disconnect_participant.go
TraceEntertains: I do have the language server enabled but I like having intellisense

Do we have like... a specific formatting preference for things in go? I know we do for ts/js with eslint but I'm not sure how that would work here

jonbarrow: Having the language server enabled is fine, I also have it enabled. It just changes what kind of formatter is used. What is the formatter you currently have set

ticket-granting/login.go
jonbarrow: Was he talking about using the server/endpoint, or was he talking about the inclusion of `v` in the semver. Because that's 2 different things. 

CC @DaniElectra for the above comment

ticket-granting/login_ex.go
jonbarrow: Sounds good, this should be noted then for the wiki at some point

match-making/database/disconnect_participant.go
jonbarrow: CC @DaniElectra I'm also getting this behavior now in https://github.com/PretendoNetwork/nex-protocols-go/pull/100. I think gofmt is doing it. We probably shouldn't fight this if that's the case

ticket-granting/login.go
DaniElectra: The library versions are all the same between endpoints, the `LibraryVersions()` function from `PRUDPEndPoint` returns the instance from the `PRUDPServer`

matchmake-extension/database/update_notification_data.go
DaniElectra: Oh well if it's gonna force it I won't fight it

match-making/database/disconnect_participant.go
DaniElectra: You should switch to an IDE that does better formatting than this /s

Oh well, I won't fight this either

DaniElectra:
> I have literally no ideas for how to avoid breaking databases so I am open to ideas :P

I don't think there's a way to avoid breaking changes, so we will just have to migrate all existing ones

match-making/database/disconnect_participant.go
jonbarrow: > You should switch to an IDE that does better formatting than this /s

I know you said /s but to be clear this is the language server/gofmt doing this, not VSC. I also tracked down the root cause, it's because gofmt will group certain operators together based on precedence. In this case the `!=` has higher precedence and so it has the spaces, but if you remove `!=` then it keeps the spaces in the bitwise operator

ticket-granting/login.go
jonbarrow: > The library versions are all the same between endpoints, the `LibraryVersions()` function from `PRUDPEndPoint` returns the instance from the `PRUDPServer`

I forgot per-endpoint configuration wasn't made yet. The intention was to add it later, since technically every endpoint can have it's own configuration (this *is* highly unlikely, and we never see this behavior used, but it's still possible). We should probably keep using the endpoint version just to future-proof things

DaniElectra: Well yes, endpoints may have their own configuration, but it's impossible for endpoints to have different versions on them. All parameters which we see as version differences are basically hardcoded in games, so this isn't a per-endpoint thing or a "configuration", but rather the NEX library as a whole (or one of the protocols) has a defined version which cannot be modified by just changing a number

jonbarrow: That's fair. The main concern I had is the fact that a single client can connect to multiple secure servers, and can send different RMC requests to each secure server. So it's possible that SecureServerA might be using a different NEX version than SecureServerB. But you're, the RMC protocols don't seem to be configurable to use different versions of themselves unlike how other systems like PRUDP can, so that would make the client incompatible

matchmake-extension/protocol.go
DaniElectra: Can you also update the `numeric(10)` on `matchmaking.gatherings` to `numeric(20)`?

TraceEntertains: Done

ticket-granting/login_ex.go
TraceEntertains: Discussed elsewhere, generally has been resolved(?)
TODO was removed regardless

ticket-granting/request_ticket.go
```diff
@@ -23,14 +26,20 @@ func (commonProtocol *CommonProtocol) requestTicket(err error, packet nex.Packet
...
 	var encryptedTicket []byte
 	if errorCode == nil {
-		encryptedTicket, errorCode = generateTicket(sourceAccount, targetAccount, commonProtocol.SessionKeyLength, endpoint)
+		encryptedTicket, errorCode = generateTicket(sourceAccount, targetAccount, nil, commonProtocol.SessionKeyLength, endpoint)
 	}
 
 	// * If any errors are triggered, return them in %retval%
 	retval := types.NewQResultSuccess(nex.ResultCodes.Core.Unknown)
 	bufResponse := types.NewBuffer(encryptedTicket)
+	pSourceKey := types.NewString("")
```
DaniElectra, line 42: I am just now noticing this, this is part of the response in this method, but it's never set here. This seems really odd, do we know if there are any instances of this method being used on user accounts, and what it returns?

ticket-granting/login_ex.go
```diff
@@ -54,6 +63,10 @@ func (commonProtocol *CommonProtocol) loginEx(err error, packet nex.PacketInterf
 		pbufResponse = types.NewBuffer(encryptedTicket)
 		strReturnMsg = commonProtocol.BuildName.Copy().(types.String)
 
+		if server.LibraryVersions.Main.GreaterOrEqual("4.0.0") && sourceKey != nil {
+			pSourceKey = types.String(fmt.Sprintf("%x", sourceKey))
```
DaniElectra: Can you use `hex.EncodeToString` for this?

ticket-granting/login.go
```diff
@@ -33,9 +34,14 @@ func (commonProtocol *CommonProtocol) login(err error, packet nex.PacketInterfac
 		targetAccount, errorCode = endpoint.AccountDetailsByUsername(commonProtocol.SecureServerAccount.Username)
 	}
 
+	if errorCode == nil && sourceAccount.RequiresTokenAuth {
+		common_globals.Logger.Error("TicketGranting::Login blocked")
```
DaniElectra: This error isn't very detailed, I would do something like `Source account requires token authentication` instead

DaniElectra (changes requested):

ticket-granting/login.go
TraceEntertains: I think keeping the "TicketGranting::Login" part is good but I just added "source account requires token authentication" to it

ticket-granting/request_ticket.go
TraceEntertains: If I recall correctly, the wiki says original servers will error on RequestTicket, but it still has the pSourceKey field

ticket-granting/login_ex.go
TraceEntertains: Sure

ticket-granting/login.go
jonbarrow: It should be noted that plogger will automatically log the package name and function name that a log is coming from. So adding those details here is a bit redundant

TraceEntertains: It's nice to have a reason without having to check the code though

The standard error for it is fine because it's kind of implied that it gets blocked due to it being explicitly disabled, but when requiring token authentication I feel it's a bit of a different case

jonbarrow: ?

This has nothing to do with the reason. I said the logger will log the package name and function calling the log. So in this case adding that is redundant. It didn't have anything to do with the log reason itself

This log will produce a log line with the text `ticket_granting.Login` already in it, because it's being logged in the `Login` function from the `ticket_granting` package. My point was that adding the `TicketGranting::Login` part is redundant in this case, since that information is already present in the package and function name

ticket-granting/request_ticket.go
DaniElectra: Oh that's right, nevermind then

ticket-granting/login.go
TraceEntertains: Oh, sorry, misinterpreted that 😭 

Will fix

ticket-granting/login.go
TraceEntertains: Did we ever get a resolution on this?

DaniElectra: Yeah I think this is resolved

ticket-granting/login.go
DaniElectra: The error log hasn't been updated yet, once that gets pushed this can be resolved

TraceEntertains: Should be updated now

DaniElectra:
Similarly to Login, the RequestTicket error log needs to be updated the same way 😅

Once that gets updated the PR *should* be finally done

DaniElectra (approved):

PR #57, opened 2025-07-26
feat(datastore): Implement notifications

DaniElectra:
Resolves #XXX

### Changes:

**NOTE: NEEDS TESTING!**

Notifications are used by HPP to share objects with specific access recipients and notify them about it. One such example is letters on Swapdoodle. Implement the necessary functionality and methods to support it.

On the database side, add a new table `datastore.notifications` with the following columns:
- `id`: The notification ID
- `read`: Whether the notification has been read with the `GetNewArrivedNotifications` method
- `recipient_id`: The target PID of the notification
- `data_id`: The referal data ID of the notification. Can be zero on notifications triggered with `GetNotificationURL`, so don't add an actual reference
- `sender_pid`: The PID who triggered the notification. For tracking purposes since this may not always be the owner (for example, on update notifications triggered by an external PID who has the update password)
- `creation_date`: When the notification was triggered. For tracking purposes
- `read_date`: If the notification has been read, timestamp of that event. For tracking purposes

Changes on the S3 side are also required, since we want to store the notification files on a separate key than object files. For this, add a separate `NotifyKeyBase` to specify the location for notification updates, as well as using a different presign function `PresignNotify` for these files for API compatibility with the original design.

Alongside that, we need a method for actually putting files into S3 from the server itself. For this, a new `PutObject` is added into `MinIOPresigner` to keep API compatibility. I don't exactly like how this is done, so I'm open to making more changes such as renaming the `S3Presigner` interface to expand its scope in name.

### DIFF
--- a/datastore/database/get_new_arrived_notifications.go
+++ b/datastore/database/get_new_arrived_notifications.go
@@ -18,21 +18,21 @@ func GetNewArrivedNotifications(manager *common_globals.DataStoreManager, recipi
 	var pResult types.List[datastore_types.DataStoreNotification]
 	var pHasNext types.Bool = false
 
-	// * Get the full count so we can determine if we are giving all notifications
-	var notificationsCount uint64
-	err = manager.Database.QueryRow(`SELECT COUNT(id) FROM datastore.notifications WHERE recipient_id = $1 AND read IS NOT TRUE`, recipientID).Scan(&notificationsCount)
+	// * Add 1 to the limit to check for more notifications remaining
+	rows, err := manager.Database.Query(`SELECT id, data_id FROM datastore.notifications WHERE recipient_id = $1 AND read IS NOT TRUE LIMIT $2+1`, recipientID, param.Limit)
 	if err != nil {
 		// TODO - Send more specific errors?
 		return nil, false, nex.NewError(nex.ResultCodes.DataStore.Unknown, err.Error())
 	}
 
-	rows, err := manager.Database.Query(`SELECT id, data_id FROM datastore.notifications WHERE recipient_id = $1 AND read IS NOT TRUE LIMIT $2`, recipientID, param.Limit)
-	if err != nil {
-		// TODO - Send more specific errors?
-		return nil, false, nex.NewError(nex.ResultCodes.DataStore.Unknown, err.Error())
-	}
+	defer rows.Close()
 
 	for rows.Next() {
+		if len(pResult) == int(param.Limit) {
+			pHasNext = true
+			break
+		}
+
 		var notification datastore_types.DataStoreNotification
 		err = rows.Scan(&notification.NotificationID, &notification.DataID)
 		if err != nil {
@@ -41,12 +41,6 @@ func GetNewArrivedNotifications(manager *common_globals.DataStoreManager, recipi
 		}
 
 		pResult = append(pResult, notification)
-	}
-
-	rows.Close()
-
-	if notificationsCount > uint64(len(pResult)) {
-		pHasNext = true
 	}
 
 	return pResult, pHasNext, nil
--- a/datastore/database/get_new_arrived_notifications.go
+++ b/datastore/database/get_new_arrived_notifications.go
@@ -1,0 +1,53 @@
+package database
+
+import (
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+	datastore_types "github.com/PretendoNetwork/nex-protocols-go/v2/datastore/types"
+)
+
+func GetNewArrivedNotifications(manager *common_globals.DataStoreManager, recipientID types.PID, param datastore_types.DataStoreGetNewArrivedNotificationsParam) (types.List[datastore_types.DataStoreNotification], types.Bool, *nex.Error) {
+	// * First we mark all unread notifications as read
+	_, err := manager.Database.Exec(`UPDATE datastore.notifications SET read = true, read_date = (CURRENT_TIMESTAMP AT TIME ZONE 'UTC') WHERE recipient_id = $1 AND id <= $2 AND read IS NOT TRUE`, recipientID, param.LastNotificationID)
+	if err != nil {
+		// TODO - Send more specific errors?
+		return nil, false, nex.NewError(nex.ResultCodes.DataStore.Unknown, err.Error())
+	}
+
+	var pResult types.List[datastore_types.DataStoreNotification]
+	var pHasNext types.Bool = false
+
+	// * Get the full count so we can determine if we are giving all notifications
+	var notificationsCount uint64
+	err = manager.Database.QueryRow(`SELECT COUNT(id) FROM datastore.notifications WHERE recipient_id = $1 AND read IS NOT TRUE`, recipientID).Scan(&notificationsCount)
+	if err != nil {
+		// TODO - Send more specific errors?
+		return nil, false, nex.NewError(nex.ResultCodes.DataStore.Unknown, err.Error())
+	}
+
+	rows, err := manager.Database.Query(`SELECT id, data_id FROM datastore.notifications WHERE recipient_id = $1 AND read IS NOT TRUE LIMIT $2`, recipientID, param.Limit)
+	if err != nil {
+		// TODO - Send more specific errors?
+		return nil, false, nex.NewError(nex.ResultCodes.DataStore.Unknown, err.Error())
+	}
+
+	for rows.Next() {
+		var notification datastore_types.DataStoreNotification
+		err = rows.Scan(&notification.NotificationID, &notification.DataID)
+		if err != nil {
+			// TODO - Send more specific errors?
+			return nil, false, nex.NewError(nex.ResultCodes.DataStore.Unknown, err.Error())
+		}
+
+		pResult = append(pResult, notification)
+	}
+
+	rows.Close()
+
+	if notificationsCount > uint64(len(pResult)) {
+		pHasNext = true
+	}
+
+	return pResult, pHasNext, nil
+}
--- a/datastore/get_new_arrived_notifications_v1.go
+++ b/datastore/get_new_arrived_notifications_v1.go
@@ -30,6 +30,7 @@ func (commonProtocol *CommonProtocol) getNewArrivedNotficationsV1(err error, pac
 	}
 
 	// * Convert types.List[datastore_types.DataStoreNotification] to types.List[datastore_types.DataStoreNotificationV1]
+	// TODO - Handle case where the data ID is bigger than the maximum supported on a uint32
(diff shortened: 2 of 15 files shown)


jonbarrow:
> Changes on the S3 side are also required, since we want to store the notification files on a separate key than object files. For this, add a separate `NotifyKeyBase` to specify the location for notification updates, as well as using a different presign function `PresignNotify` for these files for API compatibility with the original design.

We can always just be opinionated about this. We can always put objects in `KeyBase/objects` and notifications in `KeyBase/notifications` for example. That might be a cleaner solution, and would let us remove this extra function (since the `S3` struct is supposed to implement the S3 API, and these notifications aren't part of that)

DaniElectra:
That is fair yeah 👍

datastore/database/get_new_arrived_notifications.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		// TODO - Send more specific errors?
+		return nil, false, nex.NewError(nex.ResultCodes.DataStore.Unknown, err.Error())
+	}
+
+	rows, err := manager.Database.Query(`SELECT id, data_id FROM datastore.notifications WHERE recipient_id = $1 AND read IS NOT TRUE LIMIT $2`, recipientID, param.Limit)
+	if err != nil {
+		// TODO - Send more specific errors?
+		return nil, false, nex.NewError(nex.ResultCodes.DataStore.Unknown, err.Error())
+	}
```
jonbarrow: I'm no SQL expert but I feel like this can be combined into one query?

datastore/database/get_new_arrived_notifications.go
```diff
@@ -0,0 +1,53 @@
...
+		err = rows.Scan(&notification.NotificationID, &notification.DataID)
+		if err != nil {
+			// TODO - Send more specific errors?
+			return nil, false, nex.NewError(nex.ResultCodes.DataStore.Unknown, err.Error())
+		}
+
+		pResult = append(pResult, notification)
+	}
+
+	rows.Close()
```
jonbarrow: If an error occurs above, then the rows won't ever close. Usually we use `defer rows.Close()` right after the query to ensure that the rows close no matter what

datastore/database/get_new_arrived_notifications.go
```diff
@@ -0,0 +1,53 @@
...
+
+	if notificationsCount > uint64(len(pResult)) {
+		pHasNext = true
+	}
```
jonbarrow: Not really an issue but this check *can* be removed if we want to by just always setting `pHasNext` to the result of the `notificationsCount > uint64(len(pResult))` operation

datastore/get_notification_url.go
```diff
@@ -0,0 +1,59 @@
...
+		common_globals.Logger.Error(err.Error())
+		return nil, nex.NewError(nex.ResultCodes.DataStore.Unknown, "Failed to sign get request")
+	}
+
+	url := getData.URL
+
+	// * Replicate the same encoding as official servers
+	var info datastore_types.DataStoreReqGetNotificationURLInfo
+	info.URL = types.String(url.Scheme + "://" + url.Host + "/")
+	info.Key = types.String(url.Path)
```
silver-volt4: Hi, one observation from when I tested this: Path has a leading slash unless it's a relative path, which results in a pretty awkward address:

Image
(notice the double slash after `minio.pretendo.cc`)

datastore/get_new_arrived_notifications_v1.go
```diff
@@ -0,0 +1,53 @@
...
+	pResultRaw, pHasNext, errCode := database.GetNewArrivedNotifications(manager, connection.PID(), param)
+	if errCode != nil {
+		return nil, errCode
+	}
+
+	// * Convert types.List[datastore_types.DataStoreNotification] to types.List[datastore_types.DataStoreNotificationV1]
+	pResult := make(types.List[datastore_types.DataStoreNotificationV1], len(pResultRaw))
+	for i, notification := range pResultRaw {
+		pResult[i].NotificationID = notification.NotificationID
+		pResult[i].DataID = types.UInt32(notification.DataID)
```
jonbarrow, line 37: Should we maybe handle cases where DataID is larger than a uint32?

datastore/get_new_arrived_notifications_v1.go
```diff
@@ -0,0 +1,53 @@
...
+	rmcResponseStream := nex.NewByteStreamOut(endpoint.LibraryVersions(), endpoint.ByteStreamSettings())
+
+	pResult.WriteTo(rmcResponseStream)
+	pHasNext.WriteTo(rmcResponseStream)
+
+	rmcResponseBody := rmcResponseStream.Bytes()
+
+	rmcResponse := nex.NewRMCSuccess(endpoint, rmcResponseBody)
+	rmcResponse.ProtocolID = datastore.ProtocolID
+	rmcResponse.MethodID = datastore.MethodGetNewArrivedNotifications
```
jonbarrow:
```suggestion
	rmcResponse.MethodID = datastore.MethodGetNewArrivedNotificationsV1
```

datastore/protocol.go
```diff
@@ -189,6 +189,24 @@ func (commonProtocol *CommonProtocol) SetManager(manager *common_globals.DataSto
 		common_globals.Logger.Error(err.Error())
 		return
 	}
+
+	_, err = manager.Database.Exec(`CREATE TABLE datastore.notifications (
+		-- Postgres sequences of type numeric are not supported, so use bigserial instead
+		-- See: https://www.postgresql.org/docs/current/sql-createsequence.html
```
jonbarrow, line 195: Why the comment? This is pretty typical use of the type

datastore/protocol.go
```diff
@@ -189,6 +189,24 @@ func (commonProtocol *CommonProtocol) SetManager(manager *common_globals.DataSto
...
+		id bigserial PRIMARY KEY,
```
jonbarrow, line 196: If we wanted to keep it consistent we could do this the way we do it in `datastore.objects`, but this works too so not a huge deal

globals/datastore_globals.go
```diff
@@ -117,6 +126,17 @@ func (mp MinIOPresigner) PostObject(bucket, key string, lifetime time.Duration)
...
+// PutObject puts the given data into MinIO
+func (mp MinIOPresigner) PutObject(bucket, key, data string) error {
+	reader := strings.NewReader(data)
+	_, err := mp.minioClient.PutObject(context.Background(), bucket, key, reader, int64(reader.Size()), minio.PutObjectOptions{})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
```
jonbarrow: This shouldn't be in the presigner. The presigner is for creating presigned URLs to pass onto another client, whereas this method actually uploads the data to S3 directly. That makes it fundamentally not a presigner, so it should be moved somewhere else in the `S3` struct

I had originally designed this system with only presigning in mind, I did not think we would ever need to actually upload data ourselves from the server, so the current design is likely inadequate for those needs. We should probably consider restructuring the `S3` model to properly account for this

jonbarrow (changes requested):
Looks good so far, just a couple notes

datastore/get_notification_url.go
DaniElectra: Oh I see. I remember I had made a test to check the behavior of `url.Parse` and saw that it didn't include the leading slash, so I assumed it would work the same way here. Though after testing again, I'm seeing the leading slash like you say. I'll fix it

datastore/database/get_new_arrived_notifications.go
DaniElectra: The swapdoodle server PR does this in a clever way: https://github.com/SwapdoodleRevival/swapdoodle/blob/main/database/datastore/get_arrived_notifications_by_pid.go

I think I'll copy that to unify this

datastore/get_new_arrived_notifications_v1.go
DaniElectra: What should we do in that case? Throw an error?

datastore/protocol.go
DaniElectra: Also replying about why the comment: as mentioned, Postgres sequences don't support being set to `numeric`, so using the numeric type on this field would be useless since the sequence won't ever reach numbers bigger than the maximum number for `bigint` anyway.

globals/datastore_globals.go
DaniElectra: My thought was to repurpose the `S3Presigner` into a bigger scope like I'm doing here to properly account for this, so it doesn't limit itself to S3 presigning but other S3 features that may be needed like uploading data. I had kept it with the same name though since I wanted to wait for more opinions. Other suggestions for handling this are welcome

datastore/get_new_arrived_notifications_v1.go
jonbarrow: I genuinely have no idea what the best way to handle this is tbh. Maybe we should just slap a TODO comment here for later?

globals/datastore_globals.go
jonbarrow: I'm not opposed to expanding the scope, it's definitely useful. Just that the current system wasn't designed for it. So at bare minimum the "presigner" should be renamed to something more generic, or stuff moved directly into the S3 struct itself or something?

DaniElectra: I have renamed the `S3Presigner` and `MinIOPresigner` to `S3Manager` and `MinIOManager` respectively, these new names fit their purpose more

ashquarky:
@jonbarrow can we merge this

jonbarrow:
Has it been properly tested? @DaniElectra still has this marked as a draft, so until it's no longer a draft I'm hesitant to merge anything since a draft signifies the changes still aren't ready for actual use

silver-volt4:
> Has it been properly tested?

While I can't say I "properly tested" this PR, I can confirm that it works in the Swapdoodle server. I had to make a few manual fixes locally, but I believe Dani has already resolved all of the problems I came across.

ashquarky:
> Has it been properly tested? @DaniElectra still has this marked as a draft, so until it's no longer a draft I'm hesitant to merge anything since a draft signifies the changes still aren't ready for actual use

Given the merge target is your (also draft) DataStore rework branch, which also hasn't been tested a lot, I don't think it's that big of a deal? If it was dev or master sure, but it's just adding onto another PR

jonbarrow:
> > Has it been properly tested? @DaniElectra still has this marked as a draft, so until it's no longer a draft I'm hesitant to merge anything since a draft signifies the changes still aren't ready for actual use
> 
> Given the merge target is your (also draft) DataStore rework branch, which also hasn't been tested a lot, I don't think it's that big of a deal? If it was dev or master sure, but it's just adding onto another PR

That's a fair point. That's fine with me once @DaniElectra signs off that there's no more pending changes for this still, even as a draft I don't want to merge too early (it still being a draft mostly signifies to me that there's still more changes to come, so I want to make sure that isn't the case)

DaniElectra:
Functionality-wise this should be complete yeah

Issue #48, opened 2025-01-13
[Enhancement]: Expand common matchmaking implementation


shoginyan:
### Checked Existing

- [X] I have checked the repository for duplicate issues.

### What enhancement would you like to see?

`MatchmakeExtension::GetPlayingSession` needs implementing

ashquarky:
More details:
It's stubbed in Splatoon as not used, but apparently YKW3 actually does need it for successful matchmaking. Not clear why

ashquarky:
Tri Force Heroes uses GetPlayingSession also

SuperMarioDaBom:
To add clarification + additional info:

``MatchmakeExtension::GetPlayingSession`` appears to be used by the various YKW titles as a means of getting active sessions of friends, as the game makes this request pretty early on after connecting to online & the list of PIDs sent matches the console's Friend List. Not sure if this function should be restricted to just friends, or should allow a user to request any session from any PID.

Some additional items that would be nice to have implemented:
``MatchmakeExtension::GetMyBlackList (109, 24)``
``MatchmakeExtension::AddToBlackList (109, 25)``
``MatchmakeExtension::RemoveFromBlackList (109, 26)``
``MatchmakeExtension::ClearMyBlackList (109, 27)``

These are used by YKW games to kick & prevent players from rejoining rooms and/or joining a room with a player you do not wish to join. For some unknown reason, each time the game connects online for the first time since last launched, it will call ``MatchmakeExtension::ClearMyBlackList``. The game will eventually time-out if this method is not implemented. Would likely be a good idea to get this implemented so as to allow players to block each other online, even if it's technically temporary.

jonbarrow:
> Not sure if this function should be restricted to just friends, or should allow a user to request any session from any PID

The official implementation does not restrict this to only friends. The way the official servers hide users from this function is via [`MatchmakeExtension::UpdatePrivacySetting`](https://nintendo-wiki.pretendo.network/docs/nex/protocols/matchmake-extension/#23-updateprivacysetting) or by blocking the caller (if PID 1000 has blocked PID 2000, but PID 1000 has their privacy setting set to public, and then PID 2000 tries to query for PID 1000, it acts as if PID 1000 has their privacy set to private). In my testing if a user is friends with a target then the privacy setting is ignored. By default all users are set to public for both privacy settings

[...]

PR #58, opened 2025-10-20
feat(matchmake-extension): Implement blocklist functionality

Fujimon8142:
Addresses #48 ?

### Changes:

My primary motivation for implementing this was to add a block feature to "Steel Diver: Sub Wars." This PR introduces the necessary components for blocklist functionality within the MatchmakeExtension protocol.

The changes include:

- **Database Schema**: Added the `matchmaking.block_lists` table to store user block relationships.

- **Database Functions**: Implemented functions (`AddToBlockList`, `RemoveFromBlockList`, `GetBlockList`, `GetBlockedByList`) in the database package to interact with the new table.

- **RMC Handlers**: Added handlers for the `AddToBlockList`, `GetMyBlockList`, and `RemoveFromBlockList` methods. These are now registered in the common protocol.

- **Session Search Integration**: Modified `FindMatchmakeSessionBySearchCriteria` to exclude sessions hosted by blocked users or users who have blocked the searcher.

**Personal Note**:
I am not deeply familiar with the official NEX specifications or the project's internal standards for consistency and maintainability, so I'm unsure if this code fully meets requirements. To be honest, I have very little confidence.

I did test the functionality using two 3DS consoles with "Steel Diver: Sub Wars."

- [X] I have tested all of my changes.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,8 +5,8 @@ go 1.23.0
 toolchain go1.24.2
 
 require (
-	github.com/PretendoNetwork/nex-go/v2 v2.1.3
-	github.com/PretendoNetwork/nex-protocols-go/v2 v2.2.1
+	github.com/PretendoNetwork/nex-go/v2 v2.1.5
+	github.com/PretendoNetwork/nex-protocols-go/v2 v2.2.2-0.20250908132726-4a71a019af39
 	github.com/PretendoNetwork/plogger-go v1.1.0
 	github.com/PretendoNetwork/pq-extended v1.0.0
 	github.com/minio/minio-go/v7 v7.0.90
--- a/go.sum
+++ b/go.sum
@@ -1,7 +1,7 @@
-github.com/PretendoNetwork/nex-go/v2 v2.1.3 h1:hdi8PbJIWbpr3WOc1fGJ5ssF76kTxFJ3Wnz46WJYvVs=
-github.com/PretendoNetwork/nex-go/v2 v2.1.3/go.mod h1:3LyJzsv3AataJW8D0binp15Q8ZH22MWTYly1VNtXi64=
-github.com/PretendoNetwork/nex-protocols-go/v2 v2.2.1 h1:/dsuP0W7bZNvrXoXH0ZRdxpxonfbWmmson51WCQdpEQ=
-github.com/PretendoNetwork/nex-protocols-go/v2 v2.2.1/go.mod h1:+soBHmwX6ixGxj6cphLuCvfJqxcZPuowc/5e7Qi9Bz0=
+github.com/PretendoNetwork/nex-go/v2 v2.1.5 h1:7opHXanMheJi1RfL0XKhSu040XfAJlXFf/En2ZBXypk=
+github.com/PretendoNetwork/nex-go/v2 v2.1.5/go.mod h1:3LyJzsv3AataJW8D0binp15Q8ZH22MWTYly1VNtXi64=
+github.com/PretendoNetwork/nex-protocols-go/v2 v2.2.2-0.20250908132726-4a71a019af39 h1:RwXK1kf32d8KryLkjdlVMJOKAhr/FytOt9KVv6ExEpM=
+github.com/PretendoNetwork/nex-protocols-go/v2 v2.2.2-0.20250908132726-4a71a019af39/go.mod h1:+soBHmwX6ixGxj6cphLuCvfJqxcZPuowc/5e7Qi9Bz0=
 github.com/PretendoNetwork/plogger-go v1.1.0 h1:x2XgyeeM8zDFGy+NcIZd3SYC2fNrVWpBBbkqTejOfiM=
 github.com/PretendoNetwork/plogger-go v1.1.0/go.mod h1:wpltahp91IXr9nOvWgwep8zGtUKDeCVwm+/Wa484lQ4=
 github.com/PretendoNetwork/pq-extended v1.0.0 h1:GHZ0hLvCvmYKQPTV9I9XtTx8J1iB5Z9CEnfW2tUpsYg=
--- a/matchmake-extension/add_to_block_list.go
+++ b/matchmake-extension/add_to_block_list.go
@@ -1,0 +1,38 @@
+package matchmake_extension
+
+import (
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+	"github.com/PretendoNetwork/nex-protocols-common-go/v2/matchmake-extension/database"
+	matchmake_extension "github.com/PretendoNetwork/nex-protocols-go/v2/matchmake-extension"
+)
+
+func (commonProtocol *CommonProtocol) addToBlockList(err error, packet nex.PacketInterface, callID uint32, lstPrincipalID types.List[types.PID]) (*nex.RMCMessage, *nex.Error) {
+	if err != nil {
+		common_globals.Logger.Error(err.Error())
+		return nil, nex.NewError(nex.ResultCodes.Core.InvalidArgument, "change_error")
+	}
+
+	connection := packet.Sender().(*nex.PRUDPConnection)
+	endpoint := connection.Endpoint().(*nex.PRUDPEndPoint)
+
+	commonProtocol.manager.Mutex.Lock()
+	defer commonProtocol.manager.Mutex.Unlock()
+
+	nexError := database.AddToBlockList(commonProtocol.manager, connection.PID(), lstPrincipalID)
+	if nexError != nil {
+		return nil, nexError
+	}
+
+	rmcResponse := nex.NewRMCSuccess(endpoint, nil)
+	rmcResponse.ProtocolID = matchmake_extension.ProtocolID
+	rmcResponse.MethodID = matchmake_extension.MethodAddToBlockList
+	rmcResponse.CallID = callID
+
+	if commonProtocol.OnAfterAddToBlockList != nil {
+		go commonProtocol.OnAfterAddToBlockList(packet, lstPrincipalID)
+	}
+
+	return rmcResponse, nil
+}
--- a/matchmake-extension/database/add_to_block_list.go
+++ b/matchmake-extension/database/add_to_block_list.go
@@ -1,0 +1,35 @@
+package database
+
+import (
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+)
+
+// AddToBlockList adds a list of PIDs to a user's blocklist
+func AddToBlockList(manager *common_globals.MatchmakingManager, userPID types.PID, pidsToBlock types.List[types.PID]) *nex.Error {
+	tx, err := manager.Database.Begin()
+	if err != nil {
+		return nex.NewError(nex.ResultCodes.Core.Unknown, err.Error())
+	}
+
+	stmt, err := tx.Prepare(`INSERT INTO matchmaking.block_lists (user_pid, blocked_pid) VALUES ($1, $2) ON CONFLICT (user_pid, blocked_pid) DO NOTHING`)
+	if err != nil {
+		tx.Rollback()
+		return nex.NewError(nex.ResultCodes.Core.Unknown, err.Error())
+	}
+	defer stmt.Close()
+
+	for _, pidToBlock := range pidsToBlock {
+		if _, err := stmt.Exec(uint64(userPID), uint64(pidToBlock)); err != nil {
+			tx.Rollback()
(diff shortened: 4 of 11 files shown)


jonbarrow:
Thank you for the work! I'll give this a proper review soon (doing some other stuff at the moment), but I wanted to touch very quickly on this part:

> I am not deeply familiar with the official NEX specifications

The way this worked officially is that the block lists of users are *not* typically checked during match searching, and are instead checked at join time. This is typically safe, since there's no real way to know who is in a session until you join it so this can't be abused to still follow certain users. However it does mean that the client MAY attempt, and fail, to join many sessions if the user has many blocked users. It seems like Nintendo realized this limitation at some point, since they added `blockListParam` to `MatchmakeSessionSearchCriteria` in NEX 4, which presumably also filters blocked users at the search level

These are my (current) Notion notes from when I was testing things prior to the shutdown:

> ## Blocking
> 
> Users can block each other to prevent them from interacting during some in-game features. However there seems to be certain times where the user block list is not checked, and some cases where the block list *is* checked but no error is thrown.
> 
> When using search methods to find sessions by a specific user (such as with `MatchmakeExtension::GetPlayingSession`), always return empty lists when searching for players who are blocked by the caller. Do not return empty lists for searching for sessions of users the caller has blocked. No error is thrown in these cases.
> 
> When browsing for sessions in general (such as with `MatchmakeExtension::BrowseMatchmakeSession`), the block lists of both the calling user and the partipants of the session are ***not*** checked. No error is thrown in these cases. Return sessions which may have participants who are blocked by the caller, or that the caller has blocked.
> 
> When attempting to join a session (such as with `MatchmakeExtension::JoinMatchmakeSession`), the block lists of both the calling user and the partipants of the session ***are*** checked. Throw `RendezVous::DeniedByParticipants` if the caller is blocked by any participant of the session. Throw `RendezVous::ParticipantInBlackList` if the caller has blocked any participants of the session.
> 
> `MatchmakeBlockListParam` seems to have been added in NEX 4. Unknown usage. Presumably filters using the block lists at search time rather than join time.

Fujimon8142:
​Thank you for the detailed feedback! I'll try to revise the code based on your comments when I have some time.

Fujimon8142:
Thank you for the clear explanation. I understand now.
​So, just to confirm my next steps:
Should I revert the commit `feat(matchmake-extension): exclude blocked users from session search` and then implement the block check within the session join methods?
​I believe the relevant methods are `JoinMatchmakeSession`, `JoinMatchmakeSessionEx`, and `JoinMatchmakeSessionWithParam`, where I'll need to throw `ParticipantInBlackList` or `DeniedByParticipants` as you described. I'm not entirely sure which of these Join methods "Steel Diver: Sub Wars" actually calls, as I'm at school right now and can't check, but I plan to investigate when I get home.
​Does this sound like the correct approach?

jonbarrow:
Yes, ideally commit [2612175](https://github.com/PretendoNetwork/nex-protocols-common-go/pull/58/commits/2612175538d87d6adfffe8a0cca2020a84085282) should be reverted and the following changes should be made:

- `GetPlayingSession`, `GetSimplePlayingSession` and `FindCommunityByParticipant` should be updated such that gatherings are not in the returned results if the *caller* is blocked by any of the participants of the gathering. For example, if the results of these functions would normally return a gathering list such as `[ 1, 2, 3, 4 ]`, but there is a user in gathering `3` which has blocked the calling user, the data returned to the caller would be `[ 1, 2, 4 ]`. For some reason the block list does not seem to be checked in `GetDetailedParticipants`, which sort of defeats the purpose in my opinion, but it seems to be how Nintendo wanted it. That method isn't supported by us at the moment anyway, so it's fine
- None of the gathering *search* functions should check block statuses. If a gathering search such as `BrowseMatchmakeSession` would normally return a gathering that includes a user that has blocked the caller, it is ALWAYS returned anyway
- The gathering *join* methods need to check the block lists. If the caller is blocked by ANY participant of the gathering, throw `DeniedByParticipants`. If the caller has any of the participants in their block list, throw `ParticipantInBlackList`. This applies to all join methods

I know your use case is Steel Diver: Sub Wars, but when adding features to this library we do ask that they be feature-complete (within reason), so you may have to touch methods Steel Diver: Sub Wars does not use for this PR

jonbarrow:
Also, if you were interested in fully closing https://github.com/PretendoNetwork/nex-protocols-common-go/issues/48 then `UpdatePrivacySetting` would also need to be implemented. The way privacy settings work is very similar to block lists:

- If a user sets `onlineStatus` to `false`, then the results of `GetPlayingSession` and `GetSimplePlayingSession` act as if the caller is blocked by the hidden user (their results are removed from the results list)
- If a user sets `participationCommunity ` to `false`, then the results of `FindCommunityByParticipant` act as if the caller is blocked by the hidden user (their results are removed from the results list)
- All users have both settings set to `true` (visible) by default
- If a caller is friends with the hidden user, the visibility status is ignored

Fujimon8142:
Thank you for the clear instructions! I'll get to work on it when I get home.

Fujimon8142:
I have a quick question to clarify my understanding of the intended behavior for `MatchmakeExtension::GetPlayingSession`. In your previous comment, you mentioned two points that seem potentially contradictory to me regarding how it should handle cases where the caller has blocked the user being searched for:

1.  "`always return empty lists when searching for players who are blocked by the caller.`"
2.  "`Do not return empty lists for searching for sessions of users the caller has blocked.`"

Could you please clarify the correct behavior for `GetPlayingSession` specifically when the **caller** has blocked the user whose session they are trying to find? Should the session be included in the results, or should the list be empty in this case?

jonbarrow:
> I have a quick question to clarify my understanding of the intended behavior for `MatchmakeExtension::GetPlayingSession`. In your previous comment, you mentioned two points that seem potentially contradictory to me regarding how it should handle cases where the caller has blocked the user being searched for:
> 
> 1. "`always return empty lists when searching for players who are blocked by the caller.`"
> 2. "`Do not return empty lists for searching for sessions of users the caller has blocked.`"
> 
> Could you please clarify the correct behavior for `GetPlayingSession` specifically when the **caller** has blocked the user whose session they are trying to find? Should the session be included in the results, or should the list be empty in this case?

The Notion notes I sent before are somewhat old, they aren't contradictory just worded a bit poorly maybe. The gist is:

- Assume there's 2 players, PlayerA and PlayerB
- If PlayerA has blocked PlayerB, then if PlayerB tries to use `GetPlayingSession`, `GetSimplePlayingSession` or `FindCommunityByParticipant`, omit the sessions PlayerA is in from the results
- If PlayerB has blocked PlayerA, then if PlayerB tries to use `GetPlayingSession`, `GetSimplePlayingSession` or `FindCommunityByParticipant`, do not omit the sessions PlayerA is in from the results
- Notice the difference in blocking order
- None of the gathering search functions should check the block list
- Only the join methods (and probably the auto matchmake methods) should check the block list. If `dontCareMyBlackList` is set to true, then allow users to join sessions even if that session contains someone the joining user has blocked (basically, never throw `ParticipantInBlackList` if `dontCareMyBlackList` is enabled)

Fujimon8142:
Thank you again for the detailed explanation and guidance!

Based on our discussion, I've summarized my understanding of the required blocklist implementation logic. Could you please confirm if this table accurately reflects the expected behavior for the different methods? I plan to use these scenarios for testing.

**Blocklist Implementation Plan & Test Cases**

| Method Called                                      | Scenario                                       | Expected Behavior                                                                                                | Error Thrown (if any)                 | Notes                                                              |
| :------------------------------------------------- | :--------------------------------------------- | :--------------------------------------------------------------------------------------------------------------- | :------------------------------------ | :----------------------------------------------------------------- |
| **`GetPlayingSession`** <br> **`GetSimplePlayingSession`** <br> **`FindCommunityByParticipant`** | Caller (Player B) tries to find sessions of Player A, **but Player A has blocked Player B**. | Omit sessions Player A is in from the results.                                                                   | None                                  |                                                                    |
| **`GetPlayingSession`** <br> **`GetSimplePlayingSession`** <br> **`FindCommunityByParticipant`** | Caller (Player B) tries to find sessions of Player A, **and Player B has blocked Player A**.   | **Do not** omit sessions Player A is in from the results.                                                        | None                                  | Block check happens at join time.                                  |
| **`BrowseMatchmakeSession`** <br> *(Other general search methods)* | Caller (Player B) searches for sessions, and some results contain Player A who **has blocked Player B**. | Include sessions with Player A in the results.                                                                   | None                                  | Block lists are not checked during general search.                 |
| **`BrowseMatchmakeSession`** <br> *(Other general search methods)* | Caller (Player B) searches for sessions, and some results contain Player A whom **Player B has blocked**. | Include sessions with Player A in the results.                                                                   | None                                  | Block lists are not checked during general search.                 |
| **`JoinMatchmakeSession`** <br> **`JoinMatchmakeSessionEx`** <br> **`JoinMatchmakeSessionWithParam`** | Caller (Player B) tries to join a session where **a participant (Player A) has blocked Player B**. | Prevent Player B from joining the session.                                                                       | `RendezVous::DeniedByParticipants`    |                                                                    |
| **`JoinMatchmakeSession`** <br> **`JoinMatchmakeSessionEx`** <br> **`JoinMatchmakeSessionWithParam`** | Caller (Player B) tries to join a session containing Player A, **whom Player B has blocked**.    | Prevent Player B from joining the session (unless `dontCareMyBlackList` is true for `JoinMatchmakeSessionEx`). | `RendezVous::ParticipantInBlackList`  | `dontCareMyBlackList` check specific to `JoinMatchmakeSessionEx`. |
| **`AutoMatchmake...`** *(e.g., `AutoMatchmakeWithSearchCriteriaPostpone`)* | Caller (Player B) is matched into a session where **a participant (Player A) has blocked Player B**. | Prevent Player B from being placed in/joining the session.                                                       | `RendezVous::DeniedByParticipants`    | Assuming auto-matchmake performs join checks.                      |
| **`AutoMatchmake...`** *(e.g., `AutoMatchmakeWithSearchCriteriaPostpone`)* | Caller (Player B) is matched into a session containing Player A, **whom Player B has blocked**.    | Prevent Player B from being placed in/joining the session.                                                       | `RendezVous::ParticipantInBlackList`  | Assuming auto-matchmake performs join checks.                      |

Specifically regarding Steel Diver, it seems likely it uses `AutoMatchmakeWithSearchCriteriaPostpone` and doesn't explicitly call a `Join` method afterwards. Therefore, implementing the `DeniedByParticipants` and `ParticipantInBlackList` checks within the logic of `AutoMatchmakeWithSearchCriteriaPostpone` (or the internal join it might trigger) seems correct for that specific game.

Does this summary align with your expectations for the implementation? I'll proceed based on this understanding once confirmed.

jonbarrow:
Almost. That's all correct except for the auto matchmaking parts at the end. An error should never be thrown, the caller should just never be matched into a session where they are blocked by a participant or have a participant blocked. For auto matchmaking just find the first session that matches the search and has no blocklist issues, and if a session like that doesn't exist then a new one should be made like normal

Fujimon8142:
~~I am trying to set up a baseline test **using** [Yo-kai Watch Blasters](https://github.com/PretendoNetwork/yo-kai-watch-blasters). (My personal changes have not been pushed.)~~

~~My ultimate goal is to test my own modifications to `nex-protocols-common-go`, but before I do that, I need to confirm that everything works correctly with the **official, unmodified** `nex-protocols-common-go` first.~~

~~To do this, I've set up the server using the [pretendo-docker](https://github.com/MatthewL246/pretendo-docker) environment.~~

~~To clarify my testing process:~~
~~1.  I connected two consoles to my server.~~
~~2.  On the first console, I create a lobby. This triggers a method in the logs, something like `MatchmakeExtension::OpenParticipation`. (I'm at school right now, so I apologize, I can't confirm the exact method name).~~
~~3.  On the second console, I search for lobbies, which triggers `MatchmakeExtension::BrowseMatchmakeSession`.~~

~~However, even after doing this, no players (or lobbies) appear for the searching console. This happens even though I can clearly see `BrowseMatchmakeSession` appearing in the server logs.~~

~~I am having trouble diagnosing the root cause: Is this a problem with **my local [pretendo-docker](https://github.com/MatthewL246/pretendo-docker) configuration**, or is the official Pretendo implementation for [Yo-kai Watch Blasters](https://github.com/PretendoNetwork/yo-kai-watch-blasters) itself simply incomplete?~~

Never mind, I'm sorry. It seems the problem was just with my search criteria. I haven't played this game in years, I don't really remember it well...

Fujimon8142:
I was hesitant to commit untested code to this pull request, so I've been working on a separate branch ([add-block-feature-work-in-progress](https://github.com/Fujimon8142/nex-protocols-common-go/tree/add-block-feature-work-in-progress)).

In *Steel Diver: Sub Wars*, I believe `removeFromBlockList` is called during matchmaking, and this functioned correctly.

I then tried testing with *Yo-kai Watch Blasters*, but I'm running into an issue. For some reason, it seems `removeFromBlockList` is being called with `0` for the `lstPrincipalID` parameter. In this game, this function is called immediately after I unblock a user from the in-game screen, but the ID inexplicably comes through as `0`.

This is preventing me from testing the unblock functionality. Do you have any ideas what might be causing this?

Alternatively, if there are any other games that would be better for testing this feature, could you please recommend some?

PR #59, opened 2025-10-22
Tweak VacantOnly / VacantParticipants behaviour

ashquarky:
Resolves #XXX

### Changes:

Allow VacantParticipants (new field in NEX 3.4) to override the original VacantOnly boolean.

Most games do the following:
- `VacantOnly: true`
- `VacantParticipants: 1`

However, Sm4sh WiiU (and probably 3DS, I didn't check) does the following:
- `VacantOnly: false`
- `VacantParticipants: 1`

This causes it to fail matchmaking since it will always try to join sessions, even if full. However, if we give VacantParticipants precedence over VacantOnly, all is fixed. Games that (correctly) set VacantOnly to true are unaffected.

I am not aware of games that actually intentionally set `VacantOnly: false` (some kind of spectator mode?) so I was unable to see what their `VacantParticipants` value is. I hope it is 0.


ashquarky:
The other random formatting changes were done by my IDE, I guess it's stylistic

ashquarky:
@jonbarrow has informed me this is not the correct behaviour and it seems sm4sh *does* intentionally set VacantOnly to false (for some reason?)
Keeping this branch around has a hack so matchmaking can work in prod but it shouldn't merge like this

DaniElectra (commented):
While we're here we should also set all of the excludes from the search criteria to true for auto matchmake.

Something like this could work: `searchCriteria.ExcludeLocked || isAutoMatchmake`

matchmake-extension/database/find_matchmake_session_by_search_criteria.go
jonbarrow, line 219: Just a couple notes:

1. `VacantParticipants` will never be 0 in normal circumstances. When `MatchmakeSessionSearchCriteria::SetVacantOnly(bool)` is called, `VacantParticipants` is always set to 1. When `MatchmakeSessionSearchCriteria::SetVacantOnly(bool, ushort)` is called it first checks if `VacantOnly` is true or not. If it is true, then it checks the value of `VacantParticipants`, and if `VacantParticipants` is 0 then the function exits with a failure, otherwise `VacantParticipants` is updated to the given value. If `VacantOnly` is set to false, then `VacantParticipants` is always set to 1. So outside of custom clients (which I don't think should be an issue here), the `searchCriteria.VacantParticipants == 0` branch will never be hit, and tbh if a client sends 0 we should throw an error much earlier
2. The `VacantParticipants` value should still (probably) be checked in all cases, even in both auto and browse matchmaking types (which is also why it's always defaulted to 1, for 1 vacant slot for the caller). It's `VacantOnly` which we should ignore in the auto methods (can't join automatically full room), but we should still make sure there's enough spots the caller is asking for

DaniElectra:
Superseded by #67

PR #60, opened 2025-12-09
feat: Implement common utility protocol

TraceEntertains:
### Changes:
- Implements handlers for all known methods in the standard utility protocol
  - [X] AcquireNexUniqueId
  - [X] AcquireNexUniqueIdWithPassword
  - [X] AssociateNexUniqueIdWithMyPrincipalId
  - [X] AssociateNexUniqueIdsWithMyPrincipalId
  - [X] GetAssociatedNexUniqueIdWithMyPrincipalId
  - [X] GetAssociatedNexUniqueIdsWithMyPrincipalId
  - [X] GetIntegerSettings
  - [X] GetStringSettings
- Adds a UtilityManager type to common globals similar to MessagingManager and MatchmakingManager

The UtilityManager contains handlers called GenerateNEXUniqueID and GenerateNEXUniqueIDWithPassword (defaults exist and are set in utility/database/protocol.go within SetManager), and GetStringSettings/GetIntegerSettings (defaults for those do not exist and are expected to be defined by the user if required)

I have not thoroughly tested as I don't know what to test with, but I've done quite a few look overs of the code, and the logic looks solid (to me at least). I have tested the database handlers for generating unique ids and getting the associated unique ids of a user and they seem to work fine, but the point of making a pull request is mostly for peer review after all lol


globals/utility_globals.go
```diff
@@ -0,0 +1,32 @@
+package common_globals
+
+import (
+	"database/sql"
+
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+)
+
+// UtilityManager manages NEX unique ids
```
DaniElectra:
```suggestion
// UtilityManager manages NEX unique IDs
```

Please make *all* occurences of "id" uppercase

globals/utility_globals.go
```diff
@@ -0,0 +1,32 @@
...
+type UtilityManager struct {
+	Database *sql.DB
+	Endpoint *nex.PRUDPEndPoint
+
+	AllowUniqueIDStealing bool // Allows users' unique ids to be stolen while they are offline, likely unneeded but implemented for posterity
+
+	GenerateNEXUniqueID             func(manager *UtilityManager, packet nex.PacketInterface) (types.UInt64, *nex.Error)
```
DaniElectra: *Why* are we passing the whole packet interface? We should only pass what's needed from there, like the PID and such, as their own arguments

utility/acquire_nex_unique_id.go
```diff
@@ -13,19 +13,18 @@ func (commonProtocol *CommonProtocol) acquireNexUniqueID(err error, packet nex.P
...
+	pNexUniqueID, nexError := commonProtocol.manager.GenerateNEXUniqueID(commonProtocol.manager, packet)
+	if nexError != nil {
+		common_globals.Logger.Error(nexError.Error())
+		return nil, nexError
+	}
 
 	rmcResponseStream := nex.NewByteStreamOut(endpoint.LibraryVersions(), endpoint.ByteStreamSettings())
 
-	pNexUniqueID.WriteTo(rmcResponseStream)
+	types.UInt64(pNexUniqueID).WriteTo(rmcResponseStream)
```
DaniElectra: This is already a `types.UInt64`, no need to cast it

utility/acquire_nex_unique_id_with_password.go
```diff
@@ -0,0 +1,43 @@
...
+	pNexUniqueIDInfo := utility_types.NewUniqueIDInfo()
+
+	uniqueId, password, nexError := commonProtocol.manager.GenerateNEXUniqueIDWithPassword(commonProtocol.manager, packet)
+	if nexError != nil {
+		common_globals.Logger.Error(nexError.Error())
+		return nil, nexError
+	}
+
+	pNexUniqueIDInfo.NEXUniqueID = types.UInt64(uniqueId)
+	pNexUniqueIDInfo.NEXUniqueIDPassword = types.UInt64(password)
```
DaniElectra: same as before

utility/database/check_can_associate_unique_ids.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+)
+
+// This handles all relevant validity checking (including the existence of an id, if the password matches, etc)
+func CheckCanAssociateUniqueIDs(manager *common_globals.UtilityManager, userPid types.PID, uniqueIds, passwords types.List[types.UInt64]) *nex.Error {
+	var uniqueId, dbPassword types.UInt64
+	var associatedPid types.PID
```
DaniElectra:
```suggestion
	var uniqueID, dbPassword types.UInt64
	var associatedPID types.PID
```

Do this everywhere

utility/database/check_can_associate_unique_ids.go
```diff
@@ -0,0 +1,61 @@
...
+			&dbPassword,
+		)
+		if err != nil {
+			return nex.NewError(nex.ResultCodes.Core.Unknown, err.Error())
+		}
+
+		// TODO - Is this a correct assumption?
+		if associatedPid != userPid {
+			return nex.NewError(nex.ResultCodes.Core.Unknown, "One of the unique ids is already owned by this user")
+		}
```
DaniElectra: The error description doesn't match with the check, did you mean `associatedPID == userPID`?

In that case @jonbarrow can you confirm or disprove this was done by original servers?

utility/database/check_can_associate_unique_ids.go
```diff
@@ -0,0 +1,61 @@
...
+
+		targetConnection := manager.Endpoint.FindConnectionByPID(uint64(associatedPid))
```
DaniElectra: Here you want to check that `associatedPID != userPID`

utility/database/generate_nex_unique_id.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+)
+
+func GenerateNEXUniqueID(manager *common_globals.UtilityManager, packet nex.PacketInterface) (types.UInt64, *nex.Error) {
+	var uniqueID types.UInt64
+
+	err := binary.Read(rand.Reader, binary.BigEndian, &uniqueID)
```
DaniElectra:
```suggestion
	err := binary.Read(rand.Reader, binary.NativeEndian, &uniqueID)
```

little optimization

utility/database/check_user_has_primary_unique_id.go
```diff
@@ -0,0 +1,30 @@
...
+
+	err := manager.Database.QueryRow(`SELECT unique_id FROM utility.unique_ids WHERE associated_pid=$1 AND is_primary_id=true`,
+		userPid,
+	).Scan(
+		&primaryId,
+	)
+	if err != nil {
+		if err == sql.ErrNoRows {
+			primaryExists = false
+			primaryId = types.UInt64(0)
```
DaniElectra: You can assign the 0 directly

```suggestion
			primaryID = 0
```

utility/database/insert_unique_ids_by_user.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+)
+
+// isPrimary is used to indicate if the 0th unique id in the array should be set as a primary id
+func InsertUniqueIDsByUser(manager *common_globals.UtilityManager, userPid types.PID, uniqueIds types.List[types.UInt64], isPrimary bool) *nex.Error {
+	var err error
+	if len(uniqueIds) == 0 {
+		common_globals.Logger.Error("Tried to pass in empty array to InsertUniqueIDsByUser!")
+		return nex.NewError(nex.ResultCodes.Core.Unknown, "change_error")
```
DaniElectra, line 19: You can use the error description from above

utility/database/check_can_associate_unique_ids.go
```diff
@@ -0,0 +1,61 @@
...
+		if targetConnection != nil || !manager.AllowUniqueIDStealing {
```
DaniElectra:
```suggestion
		if targetConnection != nil && !manager.AllowUniqueIDStealing {
```

Otherwise this will always error

utility/database/check_can_associate_unique_ids.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"slices"
+
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+)
+
+// This handles all relevant validity checking (including the existence of an id, if the password matches, etc)
+func CheckCanAssociateUniqueIDs(manager *common_globals.UtilityManager, userPid types.PID, uniqueIds, passwords types.List[types.UInt64]) *nex.Error {
```
jonbarrow: Needs a proper [Godoc](https://go.dev/blog/godoc) comment. This goes for the whole PR

utility/database/check_can_associate_unique_ids.go
```diff
@@ -0,0 +1,61 @@
...
+	var uniqueId, dbPassword types.UInt64
```
jonbarrow: We use Google's style guide for [initialisms](https://google.github.io/styleguide/go/decisions#initialisms). This goes for the whole PR

utility/database/check_can_associate_unique_ids.go
jonbarrow: I'm gonna be so real with you I'm not sure what this error message is trying to say? The checked condition makes it seem like the error message should be something like `User does not own unique ID` or something

utility/database/generate_nex_unique_id.go
```diff
@@ -0,0 +1,34 @@
...
+	if err != nil {
+		common_globals.Logger.Error(err.Error())
+		return 0, nex.NewError(nex.ResultCodes.Core.Unknown, err.Error())
+	}
```
jonbarrow: While `crypto/rand` is cryptographically secure, it's not guaranteed to be safe from collisions. When working in the uint64 space this chance is extremely low, but not impossible. We should check if the value is in the database first before inserting

Also we can get around generating these IDs ourselves by using `gen_random_bytes` in Postgres directly, though it also does not 100% guarantee no collisions. The UUID methods/data types do, but they're 128 bits so we'd need to clamp them, which then reintroduces the collision risk

@DaniElectra thoughts?

utility/database/generate_nex_unique_id_with_password.go
```diff
@@ -0,0 +1,40 @@
...
+	if err != nil {
+		common_globals.Logger.Error(err.Error())
+		return 0, 0, nex.NewError(nex.ResultCodes.Core.Unknown, "change_error")
+	}
+
+	err = binary.Read(rand.Reader, binary.BigEndian, &password)
+	if err != nil {
+		common_globals.Logger.Error(err.Error())
+		return 0, 0, nex.NewError(nex.ResultCodes.Core.Unknown, "change_error")
+	}
```
jonbarrow: Only commenting here because I mentioned collisions before. I don't think password collisions are a big enough deal to warrant checking, to be clear

utility/database/generate_nex_unique_id_with_password.go
```diff
@@ -0,0 +1,40 @@
...
+	if nexError != nil {
+		common_globals.Logger.Error(nexError.Error())
+		return 0, 0, nexError
+	}
+
+	nexError = InsertUniqueIDsByUserWithPasswords(manager, packet.Sender().PID(), types.List[types.UInt64]{uniqueID}, types.List[types.UInt64]{password}, !primaryExists)
+	if nexError != nil {
+		common_globals.Logger.Error(nexError.Error())
+		return 0, 0, nexError
+	}
```
jonbarrow: To reduce duplication we can just use `InsertUniqueIDsByUser` and always pass in a password. For non-password IDs, just use 0 for the password

utility/database/get_user_associated_unique_ids.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+)
+
+func GetUserAssociatedUniqueIDs(manager *common_globals.UtilityManager, userPid types.PID) ([]uint64, []uint64, *nex.Error) {
+	var uniqueId, password types.UInt64
+
+	uniqueIds := make([]uint64, 0)
+	passwords := make([]uint64, 0)
```
jonbarrow: This is effectively just a deconstructed array of [`UniqueIdInfo`](https://nintendo-wiki.pretendo.network/docs/nex/protocols/utility/#uniqueidinfo-structure) types

utility/get_string_settings.go
```diff
@@ -0,0 +1,47 @@
...
+
+	if commonProtocol.manager.GetStringSettings == nil {
+		common_globals.Logger.Warning("GetStringSettings not defined")
+		return nil, nex.NewError(nex.ResultCodes.Core.NotImplemented, "change_error")
+	}
+
+	connection := packet.Sender()
+	endpoint := connection.Endpoint()
+
+	stringSettings, nexError := commonProtocol.manager.GetStringSettings(commonProtocol.manager, packet, uint32(stringSettingIndex))
```
jonbarrow: Why are you casting the value here instead of just passing everything along as-is? Consumers should be expected to work with our NEX data types

utility/get_integer_settings.go
```diff
@@ -0,0 +1,47 @@
...
+
+	if commonProtocol.manager.GetIntegerSettings == nil {
+		common_globals.Logger.Warning("GetIntegerSettings not defined")
+		return nil, nex.NewError(nex.ResultCodes.Core.NotImplemented, "change_error")
+	}
+
+	connection := packet.Sender()
+	endpoint := connection.Endpoint()
+
+	integerSettings, nexError := commonProtocol.manager.GetIntegerSettings(commonProtocol.manager, packet, uint32(integerSettingIndex))
```
jonbarrow: Why are you casting the value here instead of just passing everything along as-is? Consumers should be expected to work with our NEX data types

globals/utility_globals.go
```diff
@@ -0,0 +1,32 @@
...
+	GenerateNEXUniqueIDWithPassword func(manager *UtilityManager, packet nex.PacketInterface) (types.UInt64, types.UInt64, *nex.Error)
+	GetIntegerSettings              func(manager *UtilityManager, packet nex.PacketInterface, integerSettingIndex uint32) (map[uint16]int32, *nex.Error)
+	GetStringSettings               func(manager *UtilityManager, packet nex.PacketInterface, stringSettingIndex uint32) (map[uint16]string, *nex.Error)
```
jonbarrow: These should be expected to use our NEX data types, just like the other methods here

utility/get_associated_nex_unique_ids_with_my_principal_id.go
```diff
@@ -0,0 +1,53 @@
...
+
+	uniqueIds, passwords, nexError := utility_database.GetUserAssociatedUniqueIDs(commonProtocol.manager, packet.Sender().PID())
+	if nexError != nil {
+		common_globals.Logger.Error(nexError.Error())
+		return nil, nexError
+	}
+
+	if len(uniqueIds) == 0 || len(passwords) == 0 {
+		return nil, nex.NewError(nex.ResultCodes.Core.Unknown, "No unique ids for this user")
+	}
```
jonbarrow: Just send back an empty list?

utility/get_associated_nex_unique_ids_with_my_principal_id.go
```diff
@@ -0,0 +1,53 @@
...
+
+	uniqueIdInfos := types.NewList[utility_types.UniqueIDInfo]()
+
+	for _, i := range uniqueIds {
+		uniqueIdInfo := utility_types.NewUniqueIDInfo()
+		uniqueIdInfo.NEXUniqueID = types.UInt64(uniqueIds[i])
+		uniqueIdInfo.NEXUniqueIDPassword = types.UInt64(passwords[i])
+
+		uniqueIdInfos = append(uniqueIdInfos, uniqueIdInfo)
+	}
```
jonbarrow: Making `GetUserAssociatedUniqueIDs` create this list beforehand prevents this extra processing. The database methods are already creating lists of data, just have them return the correct type of lists

utility/associate_nex_unique_ids_with_my_principal_id.go
```diff
@@ -0,0 +1,57 @@
...
+		return nil, nex.NewError(nex.ResultCodes.Core.InvalidArgument, "No unique ids sent")
+	}
+
+	uniqueIdList := make(types.List[types.UInt64], 0)
+	passwordList := make(types.List[types.UInt64], 0)
+
+	for _, info := range uniqueIdInfos {
+		uniqueIdList = append(uniqueIdList, info.NEXUniqueID)
+		passwordList = append(passwordList, info.NEXUniqueIDPassword)
+	}
```
jonbarrow: Why do you keep using deconstructed versions of `UniqueIDInfo`? This seems like extra processing, wasting allocations and such, for no real reason, just pass around the `UniqueIDInfo` list?

utility/associate_nex_unique_id_with_my_principal_id.go
```diff
@@ -0,0 +1,49 @@
...
+		common_globals.Logger.Error(err.Error())
+		return nil, nex.NewError(nex.ResultCodes.Core.InvalidArgument, "change_error")
+	}
+
+	connection := packet.Sender()
+	endpoint := connection.Endpoint()
+
+	if uniqueIdInfo.NEXUniqueID == 0 {
+		return nil, nex.NewError(nex.ResultCodes.Core.InvalidArgument, "No unique id sent")
+	}
```
jonbarrow: Sending a value of 0 is allowed. Using 0 effectively turns this into a "remove all my IDs" call

utility/acquire_nex_unique_id.go
jonbarrow: `GenerateNEXUniqueID` already returns a `types.UInt64`, so this cast isn't needed

utility/database/update_unique_id_associations.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// isPrimary is used to indicate if the 0th unique id in the array should be set as a primary id
+func UpdateUniqueIDAssociations(manager *common_globals.UtilityManager, userPid types.PID, uniqueIds types.List[types.UInt64], isPrimary bool) *nex.Error {
+	var err error
+
+	if len(uniqueIds) == 0 {
+		common_globals.Logger.Error("Tried to pass in empty array to UpdateUniqueIDAssociationsWithPasswords!")
+		return nex.NewError(nex.ResultCodes.Core.Unknown, "change_error")
+	}
```
jonbarrow: An empty list is valid, it effectively turns `AssociateNexUniqueIdsWithMyPrincipalId` into a "clear everything" call

jonbarrow (changes requested):
Looks pretty good so far! I didn't go SUPER hard right off the bat because @DaniElectra is also reviewing and I didn't want to step on any toes, so I'm probably going to be doing a few more rounds of review after that

I didn't focus too much on the database queries in this round, mostly just the logic, since the database structure is likely to change

Also a couple of the issues I noticed are mentioned in my notes from Discord. Just to copy-paste what I said there:

> Note that this is based on research I did to `Utility`, `StorageManager` works a bit differently (though some basic ideas)
> 
> Basically it's a way to get a globally unique ID from the game server, and a NEX account can have multiple NEX unique IDs associated with it. What a game uses these IDs for is entirely up to the game, all the `Utility`/`StorageManager` protocols do is maintain what IDs exist and where they are tied
> 
> Once a unique ID is issued, a game typically ties that data to something else for later reference. The `Ranking`/`Ranking2` protocols have built in support for unique IDs too. In these protocols you can query for leaderboard data using either a PID, or a unique ID, and scores can be tied to one or the other. For example in `Ranking`, typically a user can only have 1 score per category, but if the user is assigned a unique ID then that unique ID can be used to upload another score to the category
> 
> A real world example for how you could use unique IDs in a game could be profiles. Say you want your game to have multiple profiles in a single save file, but you don't want to add the full `DataStore` protocol or extend it to support the data you need in your database. You could instead make it so each of the users game profiles is indexed using a NEX unique ID in your own database/protocol
> 
> The protocol itself is pretty simple. I don't have my detailed notes with me rn, still looking for those, but basically:
> 
> - Users can create/associate multiple NEX unique IDs to themselves
> - Users can have a "main" NEX unique ID which is used as a default value in certain contexts
> - `Utility::AcquireNexUniqueId` - Creates and returns a NEX unique ID *not* protected by a password
> - `Utility::AcquireNexUniqueIdWithPassword` - Creates and returns a NEX unique ID protected by a password
> - `Utility::AssociateNexUniqueIdWithMyPrincipalId` - Unassociates all unique IDs tied to the caller and (if the input ID is not 0 ([`INVALID_UNIQUEID`](<https://github.com/tech-ticks/RTDXTools/blob/232a7797e01369e9c12704f58fdde65dd3ac1c32/Assets/Scripts/Stubs/Generated/Assembly-CSharp/NexPlugin/Common.cs#L17>)?)) registers the input unique ID as the users "main" unique ID
> - `Utility::AssociateNexUniqueIdsWithMyPrincipalId` - Unassociates all unique IDs tied to the caller and (if the list is not empty) registers each of the input unique ID to the user. The 0th value is assigned as the "main" unique ID
> - `Utility::GetAssociatedNexUniqueIdWithMyPrincipalId` - Returns the "main" unique ID tied to a user
> - `Utility::GetAssociatedNexUniqueIdsWithMyPrincipalId` - Returns all unique IDs tied to a user. The 0th value is the "main" unique ID
> - `Utility::GetIntegerSettings` - Game-specific iirc
> - `Utility::GetStringSettings` - Game-specific iirc
> - If a user tries to associate a NEX unique ID to them and the unique ID was made with `Utility::AcquireNexUniqueIdWithPassword`, then the password MUST be used when associating
> - A unique ID cannot be associated with more than one user, but iirc it seems like you can "steal" an ID by associating an ID with your account if the previous owner is offline? I remember getting errors when trying to associate a unique ID to multiple of my accounts while both were online, but being able to do so when the original owner was offline?
> 
> An example of when a the "main" unique ID is used as the default value is in `Ranking2::GetRanking`. If the calling user has a "main" NEX unique ID associated with them, and the caller has set *both* `Ranking2GetParam.nexUniqueId` and `Ranking2GetParam.principalId` to 0, then the server uses the callers "main" unique ID to lookup rankings. If the caller has no "main" NEX unique ID, *then* the callers PID is used ("main" NEX unique ID takes priority)

utility/database/clear_pid_unique_id_associations.go
```diff
@@ -0,0 +1,17 @@
+package utility_database
+
+import (
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+)
+
+func ClearPIDUniqueIDAssociations(manager *common_globals.UtilityManager, userPid types.PID) *nex.Error {
+	_, err := manager.Database.Exec(`UPDATE utility.unique_ids SET associated_pid=0, associated_time=null, is_primary_id=false WHERE associated_pid=$1`, userPid)
```
DaniElectra: Right now when this is run these unique IDs cannot be associated back to anyone since ˋCheckCanAssociateUniqueIDsˋ doesn't account for this (the associated PID is 0 but the function doesn't check this).

I'm not sure off what the proper behavior of the associate functions are: does it allow auto-assigning unique IDs that don't exist yet or only unique IDs that were previously created? @jonbarrow Can you clarify this? The proper implementation for this will depend on how this is supposed to work.

utility/database/generate_nex_unique_id.go
DaniElectra: I'm not sure about this either, the only way I can think of to account for this is to run this on a loop like how we assign PIDs on the account server.

utility/database/check_can_associate_unique_ids.go
TraceEntertains: I was very sleep deprived when writing that file 😭

utility/database/get_user_associated_unique_ids.go
TraceEntertains: I'm aware, but I did it for a reason (to an extent), first it's because of passing in the array to postgres, its hard to check the row count (at least the easiest way I could figure out was doing it manually), there is no guarantee that a unique id will exist, so if there is an array with valid unique ids and an invalid one in the middle, you need to find the password that goes to the unique id (you can't just base it on the row index), and I just figured the easiest way to do that would be having separate arrays that I can use slices.Index on to get an index that I can use for both
Oh wait I can just make a unique ids array and do the *exact* same thing with unique id info.....

utility/database/generate_nex_unique_id_with_password.go
TraceEntertains: Works for me, tbf I think I thought about this in passing once, I just never did it lol

utility/get_string_settings.go
TraceEntertains: I was working off the dated assumption that they aren't (most user-defined handlers have used native types in my experience)

utility/database/check_can_associate_unique_ids.go
TraceEntertains: Just want to check before I go make any changes, would correct names for `userPid`, `uniqueIds`, and `uniqueId` be `userPID`, `uniqueIDs`, and `uniqueID` respectively?

utility/database/update_unique_id_associations.go
```diff
@@ -0,0 +1,41 @@
...
+
+	currentTime := time.Now().UTC()
+
+	for index, uniqueId := range uniqueIds {
+		_, err = manager.Database.Exec(`UPDATE utility.unique_ids SET 
+				associated_pid=$2, 
+				associated_time=$3, 
+				is_primary_id=$4
+			WHERE unique_id=$1
```
DaniElectra: This can probably be optimized as a single query (without the for loop) by using `WHERE unique_id=ANY($)` and determining the index for `is_primary_id` inside the query:

```sql
UPDATE utility.unique_ids SET 
	associated_pid=$2, 
	associated_time=$3, 
	is_primary_id=(CASE WHEN unique_id=$5 AND $6 THEN true ELSE false) -- $5 would be the uniqueIDs[0] and $6 would be isPrimary
WHERE unique_id=ANY($1) -- $1 would be the uniqueIDs
```

utility/get_integer_settings.go
```diff
@@ -0,0 +1,47 @@
...
+		common_globals.Logger.Error(nexError.Error())
+		return nil, nexError
+	}
+
+	rmcResponseStream := nex.NewByteStreamOut(endpoint.LibraryVersions(), endpoint.ByteStreamSettings())
+
+	nexIntegerSettings := make(types.Map[types.UInt16, types.Int32])
+	for k, v := range integerSettings {
+		nexIntegerSettings[types.UInt16(k)] = types.Int32(v)
+	}
```
DaniElectra: Similarly to what Jon said, the function above can just return a NEX type `types.Map` so we don't have to do this conversion

utility/database/check_can_associate_unique_ids.go
DaniElectra: Yes that's correct

utility/get_string_settings.go
DaniElectra: While it's true that most of match-making uses native types, that is for historical reasons: the common matchmaking was developed before the revamp on types that happened on nex-go v2.1.0.

Since then we have now started using our types on these implementations (see common messaging as an example)

DaniElectra (changes requested):
Most of these are duplicates of what Jon has said, but I can't bother to remove them now lmao

utility/database/check_can_associate_unique_ids.go
jonbarrow: I seem to have lost my hyper detailed notes at some point when moving computers, but;

If you're asking about how this works given the current check (`associatedPid != userPid`):

- If an ID is assigned to UserA, and UserA is online, the ID cannot be associated with UserB
- If an ID is assigned to UserA, and UserA is offline, UserB can associate the ID with themselves and it is removed from UserA

I do not remember what the error code was though, and to be quite honest I cannot think of a legitimate reason for this feature. It sounds somewhat dangerous and I'm not sure if it's even intended behvior (I very well could have found a bug in the protocol), so I'm not sure if we should implement it unless we see a game using this feature

If you're asking about your proposed check (`associatedPID == userPID`):

Trying to assign an ID to yourself that you already have assigned is intended usage. `AssociateNexUniqueIdsWithMyPrincipalId` will first wipe out all previous IDs and then assign the ones provided by the client, so if a client needs to preserve any IDs it would need to pass them to the server (effectively "reassigning" the IDs to themselves). Otherwise the client would lose IDs it needs

For example, if a client already has IDs 100, 200, and 300 assigned to themselves, but now wants to *remove* ***only*** ID 200, it would call `AssociateNexUniqueIdsWithMyPrincipalId([ 100, 300 ])` and ID 100 would be made the users "main" ID

In these cases, when an ID is removed from a user typically all data is also removed. From my `Ranking2` notes:

```
- If using `Utility::AssociateNexUniqueIdWithMyPrincipalId` (singular), ALL other associated unique IDs, and their data, are DELETED!
- If using `Utility::AssociateNexUniqueIdsWithMyPrincipalId` (multiple), ALL unique IDs that are already associated with the user and are NOT sent are DELETED, along with their data. Must send ALL unique IDs, even ones previously associated!
```

However I did neglect to check if an ID can then be *reassigned* to a new user after being removed from it's previous owner

utility/database/clear_pid_unique_id_associations.go
jonbarrow: > Right now when this is run these unique IDs cannot be associated back to anyone

In my testing I believe I neglected to check if it was possible to reassign an ID after it was cleared from a previous user. I may setup my Switch to check this though, and just risk it

> does it allow auto-assigning unique IDs that don't exist yet or only unique IDs that were previously created?

When an ID is created it seems to just be totally random, I have no idea if it was selecting from a pool that includes previously made IDs but I doubt it. Wouldn't be much reason to do so imo, I'm fairly certain it's just entirely random IDs. The IDs get auto-assigned to the creator though, iirc

utility/database/check_can_associate_unique_ids.go
TraceEntertains: This should be resolved in the latest commit, will leave conversation open in case of any objections

utility/database/check_can_associate_unique_ids.go
TraceEntertains: This should be resolved in the latest commit, will leave conversation open in case of any objections

utility/database/check_can_associate_unique_ids.go
TraceEntertains: Resolved in the latest commit by removing the check entirely, I realized that it shouldn't matter anyways considering it only gets called in methods that unassociate all other unique IDs from the user

utility/database/generate_nex_unique_id.go
TraceEntertains: I made a function to check if it exists in the database already, then I made the very questionable decision of calling it from itself in the instance that it does exist already 😭

Leaving conversation open in case anyone has a better suggestion that isn't too complicated

utility/database/generate_nex_unique_id_with_password.go
TraceEntertains: It's not like passwords are a primary key so I'm not too concerned about it tbf

utility/database/generate_nex_unique_id_with_password.go
TraceEntertains: Done in latest commit

utility/get_string_settings.go
TraceEntertains: Fixed in latest commit

globals/utility_globals.go
TraceEntertains: Fixed in latest commit

utility/get_integer_settings.go
TraceEntertains: Fixed in latest commit

utility/get_associated_nex_unique_ids_with_my_principal_id.go
TraceEntertains: Fixed in latest commit

utility/get_associated_nex_unique_ids_with_my_principal_id.go
TraceEntertains: Fixed in latest commit

utility/associate_nex_unique_id_with_my_principal_id.go
TraceEntertains: Fixed in latest commit

utility/acquire_nex_unique_id.go
TraceEntertains: Fixed in latest commit

utility/database/update_unique_id_associations.go
TraceEntertains: Fixed in latest commit

globals/utility_globals.go
TraceEntertains: Fixed in latest commit (at least, I hope it is 😭)

globals/utility_globals.go
TraceEntertains: Fixed in latest commit

utility/acquire_nex_unique_id.go
TraceEntertains: Fixed in latest commit

utility/acquire_nex_unique_id_with_password.go
TraceEntertains: Fixed in latest commit

utility/database/check_can_associate_unique_ids.go
TraceEntertains: Fixed in latest commit

utility/database/check_can_associate_unique_ids.go
TraceEntertains: Added to the statement that checks for unique ID stealing/if owner is online

utility/database/check_user_has_primary_unique_id.go
TraceEntertains: Fixed in latest commit

utility/database/generate_nex_unique_id.go
TraceEntertains: Applied in latest commit

utility/database/check_can_associate_unique_ids.go
TraceEntertains: Like the comment I left on Jon's review, solved, pretty much just removed it because it was redundant iirc

utility/database/insert_unique_ids_by_user.go
TraceEntertains: You sure....? Both the passwords and uniqueIDs array can (\*\*\*\*\*theoretically, if done by a developer directly using the database function with improper checks) be 0 in length, which this would catch

utility/database/check_can_associate_unique_ids.go
TraceEntertains: It shouldn't
As of latest commit, when unique id stealing is enabled, if the user is not the owner, and the owner is online, it will error
When unique id stealing is disabled, and the user is not the owner, it will error

utility/database/clear_pid_unique_id_associations.go
TraceEntertains: Oh, I didn't think about that until now
Fixed in latest commit

utility/database/update_unique_id_associations.go
TraceEntertains: Done

utility/get_integer_settings.go
TraceEntertains: Done

utility/associate_nex_unique_ids_with_my_principal_id.go
TraceEntertains: Most instances of this have been fixed except ones where it's difficult to do it any other way (i.e. lists of unique IDs to pass in to the db)

utility/database/get_user_associated_unique_ids.go
jonbarrow: I'm not sure I understand what you mean, you don't have to pass the array to postgres or use the index or anything like that? Just make a list of `UniqueIdInfo` and use that. Here's a corrected example, which also includes the missing `defer rows.Close()` (this is important, I missed it earlier, without calling `defer rows.Close()` this will cause a memory leak):

```go
package utility_database

import (
	"github.com/PretendoNetwork/nex-go/v2"
	"github.com/PretendoNetwork/nex-go/v2/types"
	utility_types "github.com/PretendoNetwork/nex-protocols-go/v2/utility/types"
	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
)

func GetUserAssociatedUniqueIDs(manager *common_globals.UtilityManager, userPID types.PID) (types.List[utility_types.UniqueIDInfo], *nex.Error) {
	uniqueIDInfos := types.NewList[utility_types.UniqueIDInfo]()

	rows, err := manager.Database.Query(`
			SELECT
				unique_id,
				password
			FROM utility.unique_ids
			WHERE associated_pid=$1 ORDER BY is_primary_id DESC, associated_time DESC`,
		userPID,
	)
	if err != nil {
		return nil, nex.NewError(nex.ResultCodes.Core.Unknown, err.Error())
	}
	defer rows.Close()

	for rows.Next() {
		uniqueIDInfo := utility_types.NewUniqueIDInfo()

		err = rows.Scan(
			&uniqueIDInfo.NEXUniqueID,
			&uniqueIDInfo.NEXUniqueIDPassword,
		)
		if err != nil {
			return nil, nex.NewError(nex.ResultCodes.Core.Unknown, err.Error())
		}

		uniqueIDInfos = append(uniqueIDInfos, uniqueIDInfo)
	}

	return uniqueIDInfos, nil
}
```

TraceEntertains: Jon that's pretty much what the code is right now 😭
I was talking in regards to primarily CheckCanAssociateUniqueIDs, UpdateUniqueIDAssociations, and (to an extent) InsertUniqueIDsByUser, so that's my bad bc I didn't check where I was posting my comment at first lol

TraceEntertains: Will add the `defer rows.Close()`, didn't even think of that

jonbarrow: > Jon that's pretty much what the code is right now

No, it's not.

The current implementation creates 3 + N number of allocations (2 slice allocations in `GetUserAssociatedUniqueIDs`, 1 slice allocation in the caller (such as `getAssociatedNEXUniqueIDsWithMyPrincipalID`) and then N object allocations for the number of `UniqueIDInfo` structs created), and holds more data in memory until garbage collected due to this. It also requires converting the deconstructed slices back into the `List[UniqueIDInfo]` anyway for actual use (the only 2 callers of this function both convert it), which both wastes time/cycles and creates additional code complexity.

The example I gave creates only 1 + N number of allocations (1 slice allocation and then N object allocations for the number of `UniqueIDInfo` structs created in `GetUserAssociatedUniqueIDs`), and removes the need to reconstruct the `List[UniqueIDInfo]` thus reducing code complexity.

Just use the correct data structures from the get-go and you free up any issues like this and make the code easier to navigate about.

TraceEntertains: Idk what you're seeing then, this is literally what's pushed right now:
image

jonbarrow: It's showing as completely unchanged for me

TraceEntertains: Have you tried clearing your cache/restarting your browser or something 😭

utility/database/clear_pid_unique_id_associations.go
TraceEntertains: > > Right now when this is run these unique IDs cannot be associated back to anyone
> 
> In my testing I believe I neglected to check if it was possible to reassign an ID after it was cleared from a previous user. I may setup my Switch to check this though, and just risk it
> 
> > does it allow auto-assigning unique IDs that don't exist yet or only unique IDs that were previously created?
> 
> When an ID is created it seems to just be totally random, I have no idea if it was selecting from a pool that includes previously made IDs but I doubt it. Wouldn't be much reason to do so imo, I'm fairly certain it's just entirely random IDs. The IDs get auto-assigned to the creator though, iirc

Jon if you risk your switch can you also check some stuff on subscription

utility/acquire_nex_unique_id.go
```diff
@@ -2,12 +2,11 @@ package utility
 
 import (
 	"github.com/PretendoNetwork/nex-go/v2"
-	"github.com/PretendoNetwork/nex-go/v2/types"
 	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
 	utility "github.com/PretendoNetwork/nex-protocols-go/v2/utility"
 )
 
-func (commonProtocol *CommonProtocol) acquireNexUniqueID(err error, packet nex.PacketInterface, callID uint32) (*nex.RMCMessage, *nex.Error) {
+func (commonProtocol *CommonProtocol) acquireNEXUniqueID(err error, packet nex.PacketInterface, callID uint32) (*nex.RMCMessage, *nex.Error) {
```
DaniElectra: Sorry I did not mean *literally everything*, we should keep the protocol method handlers the same as their name on the protocols library

utility/acquire_nex_unique_id_with_password.go
```diff
@@ -17,20 +15,15 @@ func (commonProtocol *CommonProtocol) acquireNexUniqueIDWithPassword(err error,
...
 		return nil, nexError
 	}
 
-	pNexUniqueIDInfo.NEXUniqueID = types.UInt64(uniqueId)
-	pNexUniqueIDInfo.NEXUniqueIDPassword = types.UInt64(password)
-
 	rmcResponseStream := nex.NewByteStreamOut(endpoint.LibraryVersions(), endpoint.ByteStreamSettings())
 
-	pNexUniqueIDInfo.WriteTo(rmcResponseStream)
+	pUniqueIDInfo.WriteTo(rmcResponseStream)
```
DaniElectra: Let's preserve the original name here

```suggestion
	pNexUniqueIDInfo.WriteTo(rmcResponseStream)
```

utility/database/check_can_associate_unique_ids.go
```diff
@@ -6,55 +6,61 @@ import (
...
-// This handles all relevant validity checking (including the existence of an id, if the password matches, etc)
-func CheckCanAssociateUniqueIDs(manager *common_globals.UtilityManager, userPid types.PID, uniqueIds, passwords types.List[types.UInt64]) *nex.Error {
-	var uniqueId, dbPassword types.UInt64
-	var associatedPid types.PID
+// CheckCanAssociateUniqueIDs handles all relevant validity checking (including the existence of a unique ID, if the password matches, etc)
+func CheckCanAssociateUniqueIDs(manager *common_globals.UtilityManager, userPID types.PID, uniqueIDInfos types.List[utility_types.UniqueIDInfo]) *nex.Error {
+	var uniqueID, dbPassword types.UInt64
+	var associatedPID types.PID
+
+	uniqueIDs := make([]types.UInt64, 0)
```
DaniElectra: You want to use our `types.List` type, otherwise the database query won't work out of the box

```suggestion
	uniqueIDs := make(types.List[types.UInt64], 0)
```

utility/database/check_unique_id_already_exists.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+	utility_types "github.com/PretendoNetwork/nex-protocols-go/v2/utility/types"
+)
+
+// CheckUniqueIDAlreadyExists takes a UniqueIDInfo and returns an error if the unique ID is already present in the database
+func CheckUniqueIDAlreadyExists(manager *common_globals.UtilityManager, uniqueIDInfo utility_types.UniqueIDInfo) *nex.Error {
+	var primaryID types.UInt64
```
DaniElectra:
```suggestion
	var uniqueID types.UInt64
```

utility/database/check_can_associate_unique_ids.go
```diff
@@ -0,0 +1,67 @@
...
+		)
+		if err != nil {
+			return nex.NewError(nex.ResultCodes.Core.Unknown, err.Error())
+		}
+
+		var targetConnection *nex.PRUDPConnection = nil
+		if associatedPID != 0 {
+			targetConnection = manager.Endpoint.FindConnectionByPID(uint64(associatedPID))
+
+			if !manager.AllowUniqueIDStealing && associatedPID != userPID || targetConnection != nil && associatedPID != userPID {
```
DaniElectra:
```suggestion
		if associatedPID != 0 && associatedPID != userPID {
			targetConnection = manager.Endpoint.FindConnectionByPID(uint64(associatedPID))

			if !manager.AllowUniqueIDStealing || targetConnection != nil {
```

globals/utility_globals.go
DaniElectra: You didn't change this one, so definitely not all occurrences have been fixed 😅

DaniElectra (changes requested):

utility/acquire_nex_unique_id.go
jonbarrow: This is actually consistent with our other implementations. Like in `SecureConnection`, the real name is `RequestUrls` but we use the function name `requestURLs`. We've typically changed real names to match the Go standard

DaniElectra: Yes, but on NEX-protocols-go this is currently named `AcquireNexUniqueID` so I'd say to keep it that way instead of making NEX all uppercase

jonbarrow: In that case I agree with keeping it consistent with the current capitalization, we should probably circle back around to this though at some point in the future to see about making NEX consistent with our other naming though. It looks like that got missed (we use the capitalization "NEX" in other areas, like `UniqueIDInfo.NEXUniqueID`)

TraceEntertains: Changed to match protocols-go names for consistency

jonbarrow (commented):
I haven’t done a full review the latest commits yet because I’m not home right now, but I just got an email about this commit.

Please do more frequent, scoped, commits with actual descriptive commit messages, rather than a single commit that covers multiple changes and a non-descriptive message. “Second set of suggestions” does not give any indication what your changes actually are. There is an entire section of our contribution guide about this, and I’ve mentioned it a number of times before.

utility/associate_nex_unique_ids_with_my_principal_id.go
```diff
@@ -0,0 +1,50 @@
...
+		return nil, nexError
+	}
+
+	if infosLength != 0 {
+		nexError = utility_database.UpdateUniqueIDAssociations(commonProtocol.manager, packet.Sender().PID(), uniqueIDInfos, true)
+		if nexError != nil {
+			common_globals.Logger.Error(nexError.Error())
+			return nil, nexError
+		}
+	}
```
jonbarrow: This can be simplified down to just:


```suggestion
	nexError := utility_database.ClearPIDUniqueIDAssociations(commonProtocol.manager, packet.Sender().PID())
	if nexError != nil {
		common_globals.Logger.Error(nexError.Error())
		return nil, nexError
	}

	nexError = utility_database.UpdateUniqueIDAssociations(commonProtocol.manager, packet.Sender().PID(), uniqueIDInfos, true)
	if nexError != nil {
		common_globals.Logger.Error(nexError.Error())
		return nil, nexError
	}
```

By just checking `CheckCanAssociateUniqueIDs` and the slice length inside `UpdateUniqueIDAssociations`. This also helps reduce the duplicated logic in `associateNexUniqueIDWithMyPrincipalID`

utility/associate_nex_unique_id_with_my_principal_id.go
```diff
@@ -0,0 +1,49 @@
...
+		return nil, nexError
+	}
+
+	if uniqueIDInfo.NEXUniqueID != 0 {
+		nexError = utility_database.UpdateUniqueIDAssociations(commonProtocol.manager, packet.Sender().PID(), types.List[utility_types.UniqueIDInfo]{uniqueIDInfo}, true)
+		if nexError != nil {
+			common_globals.Logger.Error(nexError.Error())
+			return nil, nexError
+		}
+	}
```
jonbarrow: This can also be simplified in a similar way to `associateNexUniqueIDsWithMyPrincipalID`

utility/get_associated_nex_unique_ids_with_my_principal_id.go
```diff
@@ -0,0 +1,46 @@
...
+		common_globals.Logger.Error(nexError.Error())
+		return nil, nexError
+	}
+
+	rmcResponseStream := nex.NewByteStreamOut(endpoint.LibraryVersions(), endpoint.ByteStreamSettings())
+
+	if len(uniqueIDInfos) == 0 {
+		uniqueIDInfo := utility_types.NewUniqueIDInfo()
+		uniqueIDInfo.NEXUniqueID = 0
+		uniqueIDInfo.NEXUniqueIDPassword = 0
```
jonbarrow: `NewUniqueIDInfo` already sets these values to 0

utility/get_associated_nex_unique_id_with_my_principal_id.go
```diff
@@ -0,0 +1,46 @@
...
+		common_globals.Logger.Error(nexError.Error())
+		return nil, nexError
+	}
+
+	rmcResponseStream := nex.NewByteStreamOut(endpoint.LibraryVersions(), endpoint.ByteStreamSettings())
+
+	if len(uniqueIDInfos) == 0 {
+		uniqueIDInfo := utility_types.NewUniqueIDInfo()
+		uniqueIDInfo.NEXUniqueID = 0
+		uniqueIDInfo.NEXUniqueIDPassword = 0
```
jonbarrow: `NewUniqueIDInfo` already sets these values to 0

utility/database/generate_nex_unique_id_with_password.go
```diff
@@ -0,0 +1,51 @@
...
+
+	nexError = InsertUniqueIDsByUser(manager, userPID, types.List[utility_types.UniqueIDInfo]{uniqueIDInfo}, !primaryExists)
+	if nexError != nil {
+		common_globals.Logger.Error(nexError.Error())
+		return utility_types.UniqueIDInfo{}, nexError
+	}
+
+	return uniqueIDInfo, nil
+}
```
jonbarrow: This can be simplified. This logic is mostly duplicated in `GenerateNEXUniqueID`, the only difference being that it also generates a password. Since passwords don't strictly need to be unique, we can dog-food a little here to remove the duplicate code. `GenerateNEXUniqueIDWithPassword` can just call `GenerateNEXUniqueID` internally, generate the password, and then update the row with the password

jonbarrow (changes requested):

utility/database/generate_nex_unique_id_with_password.go
TraceEntertains: Tbf, running an insert statement just to run an update statement after feels kinda yucky, so I made GenerateNEXUniqueIDWithPassword the default and made generating a password into a flag, as they just use the same InsertUniqueIDsByUser function anyways, with the key difference being the password, which GenerateNEXUniqueID set to zero anyways (same as what would happen when passing in a new UniqueIDInfo object with only a unique ID set)

jonbarrow (commented):
LGTM. Giving this a tentative approval, pending review from @DaniElectra to make sure I didn't miss anything

DaniElectra (dismissed):
LGTM

PR #61, opened 2026-01-17
Use gRPC server to validate NEX token instead of local validation

jonbarrow:
Resolves #XXX

### Changes:

Makes the following changes:

- Uses gRPC to contact the account server to validate NEX tokens, instead of manually parsing them. This helps future-proof the servers in case of token changes and lets us revoke tokens and such
- Renames `ValidatePretendoLoginData` to `ValidateNEXLoginData` for clarity on what kind of data is being validated (and also removes the branding issue of not having "Network" in the name, without making the function name too long)
- Renames `SetPretendoValidation` to `SetGameServerID` for clarity on what the value is that's actually being set. The game server ID is tied to tokens on the account server, so sending it allows the account server to know that the token was issued for the requesting game server
    - Note 1: The current account server implementation does not check the game server ID just to make things simple for now, however this will likely come sooner than later once we implement things like per-game bans
    - Note 2: I did mess up a bit however and made it so the gRPC server can only recieve one game server ID and the NEX token exchange endpoint doesn't allow the client to send title IDs. This means certain games which share a game server, but not necessarily the same game server ID (such as Pokemon) will have issues unless we either split the server, change the gRPC server to allow title IDs, or change the gRPC server to allow multiple game server IDs (either by changing the protocols or by allowing the single field to contain multiple game server IDs, like `ID1,ID2,ID3` and the server splits them). Changing the gRPC protobufs to allow multiple game server IDs is the cleanest option imo, and I would prefer making it a real array field, but keeping it as a single string that gets split on the server is also fine and would mean less API changes

Do not merge this yet even if approved. Needs https://github.com/PretendoNetwork/account/pull/313 to be both merged AND deployed to work


globals/authentication.go
```diff
@@ -106,7 +134,7 @@ func DecryptToken(encryptedToken []byte, aesKey []byte) (*NEXToken, *nex.Error)
 	expectedChecksum := binary.BigEndian.Uint32(encryptedToken[0:4])
 	encryptedBody := encryptedToken[4:]
 
-	if len(encryptedBody) % aes.BlockSize != 0 {
```
DaniElectra, line 109: With the new authentication this `DecryptToken` function can be removed

globals/authentication.go
```diff
@@ -2,23 +2,47 @@ package common_globals
...
+	}
+
+	GRPCAccountClient = pb.NewAccountServiceClient(GRPCAccountClientConnection)
+	GRPCAccountCommonMetadata = metadata.Pairs(
+		"X-API-Key", apiKey,
+	)
+}
+
+// ValidateNEXLoginData validates the given NEX login data
+func ValidateNEXLoginData(pid types.PID, loginData types.DataHolder, gameServerID string) *nex.Error {
```
DaniElectra: I'd prefer a name that clearly shows that this is our own validation to avoid confusion, such as `ValidatePNLoginData` which would respect our branding

secure-connection/protocol.go
```diff
@@ -32,10 +32,10 @@ func (commonProtocol *CommonProtocol) EnableInsecureRegister() {
 	commonProtocol.allowInsecureRegisterMethod = true
 }
 
-// SetPretendoValidation configures the protocol to use Pretendo validation
-func (commonProtocol *CommonProtocol) SetPretendoValidation(aesKey []byte) {
+// SetGameServerID configures the servers game server ID for use in validating NEX login data
+func (commonProtocol *CommonProtocol) SetGameServerID(gameServerID string) {
```
DaniElectra: Similar to the above, I'd prefer a name that shows this is for an opinionated implementation. I'm thinking on `ConfigurePNValidation`, but other options are welcome

DaniElectra (changes requested):

globals/authentication.go
```diff
@@ -90,54 +113,3 @@ type NEXToken struct {
 	TitleID     uint64
 	AccessLevel int8
 }
```
DaniElectra: I think this `NEXToken` struct can also be removed

globals/authentication.go
```diff
@@ -1,24 +1,43 @@
...
 	account_management_types "github.com/PretendoNetwork/nex-protocols-go/v2/account-management/types"
 	ticket_granting_types "github.com/PretendoNetwork/nex-protocols-go/v2/ticket-granting/types"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/metadata"
 )
 
-// ValidatePretendoLoginData validates the given Pretendo login data
-func ValidatePretendoLoginData(pid types.PID, loginData types.DataHolder, aesKey []byte) *nex.Error {
+func ConnectToAccountGRPC(host string, port uint16, apiKey string) {
```
DaniElectra, line 22:
```suggestion
// ConnectToAccountGRPC connects to Pretendo Network's account server GRPC
func ConnectToAccountGRPC(host string, port uint16, apiKey string) {
```

jonbarrow: I made the Godoc comment a bit more generic since other people can use it to connect to their own servers too

PR #62, opened 2026-04-13
Update to protocol libs v2.3

jonbarrow:
Resolves #XXX

### Changes:

Updates to protocol libs v2.3.X (currently only 2.3.0 https://github.com/PretendoNetwork/nex-protocols-go/releases/tag/v2.3.0)

Relies on unmerged changes:

- https://github.com/PretendoNetwork/nex-protocols-go/pull/109
- https://github.com/PretendoNetwork/nex-protocols-go/pull/110

Marking as draft for now since it's still broken. @DaniElectra the only stuff left here is the database stuff I think? I don't want to step on your toes though and break anything since you were the one who designed the messaging/matchmaking database stuff


jonbarrow:
@DaniElectra This is ready for review now. I have not tested everything to know if it all works, but it's just type changes so it should work just fine. Only thing that may be funky is the database stuff? Unsure if anything else is needed there. We implement all the SQL scanner and such interfaces in `nex-go`, which the constants build off of, so I ***THINK*** it should be fine?

DaniElectra:
I believe that is fine yeah. Though I have just noticed that we might have issues with the enum types on structs since they don't have a `String()` method but still use `%s` on string formatting on structs. For example: https://github.com/PretendoNetwork/nex-protocols-go/blob/c23c5b38eb184b103fa540a45fe829e0a5b62695/match-making/types/gathering.go#L212

We probably want to add `String()` methods to the enum types so that struct printing works properly

jonbarrow:
@DaniElectra `String()` methods have been added 👍

DaniElectra (approved):
There are some constants which remain to be updated such as the recipient type in some places, but I don't want to make it a blocker for other work so LGTM, we can replace the remaining bits later

PR #65, opened 2026-07-19
fix: make passwords behave correctly

RusticMaple:
### Changes:
- Fixes not correctly respecting wether or not the matchmake password is set
- Fixes ways of bypassing passwords

### DIFF
--- a/matchmake-extension/database/find_matchmake_session_by_search_criteria.go
+++ b/matchmake-extension/database/find_matchmake_session_by_search_criteria.go
@@ -16,7 +16,7 @@ import (
 )
 
 // FindMatchmakeSessionBySearchCriteria finds matchmake sessions with the given search criterias
-func FindMatchmakeSessionBySearchCriteria(manager *common_globals.MatchmakingManager, connection *nex.PRUDPConnection, searchCriterias []match_making_types.MatchmakeSessionSearchCriteria, resultRange types.ResultRange, matchmake bool, sourceMatchmakeSession *match_making_types.MatchmakeSession) ([]match_making_types.MatchmakeSession, *nex.Error) {
+func FindMatchmakeSessionBySearchCriteria(manager *common_globals.MatchmakingManager, connection *nex.PRUDPConnection, searchCriterias []match_making_types.MatchmakeSessionSearchCriteria, resultRange types.ResultRange, publicSession bool, sourceMatchmakeSession *match_making_types.MatchmakeSession) ([]match_making_types.MatchmakeSession, *nex.Error) {
 	resultMatchmakeSessions := make([]match_making_types.MatchmakeSession, 0)
 
 	endpoint := connection.Endpoint().(*nex.PRUDPEndPoint)
@@ -208,7 +208,7 @@ func FindMatchmakeSessionBySearchCriteria(manager *common_globals.MatchmakingMan
 			}
 		}
 
-		if matchmake {
+		if publicSession {
 			searchStatement += ` AND ms.user_password="" AND ms.system_password=""`
 		}
 
--- a/matchmake-extension/database/find_matchmake_session_by_search_criteria.go
+++ b/matchmake-extension/database/find_matchmake_session_by_search_criteria.go
@@ -16,7 +16,7 @@ import (
 )
 
 // FindMatchmakeSessionBySearchCriteria finds matchmake sessions with the given search criterias
-func FindMatchmakeSessionBySearchCriteria(manager *common_globals.MatchmakingManager, connection *nex.PRUDPConnection, searchCriterias []match_making_types.MatchmakeSessionSearchCriteria, resultRange types.ResultRange, sourceMatchmakeSession *match_making_types.MatchmakeSession) ([]match_making_types.MatchmakeSession, *nex.Error) {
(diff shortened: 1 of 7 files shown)


matchmake-extension/database/find_matchmake_session_by_search_criteria.go
```diff
@@ -208,6 +208,10 @@ func FindMatchmakeSessionBySearchCriteria(manager *common_globals.MatchmakingMan
 			}
 		}
 
+		if matchmake {
+			searchStatement += ` AND ms.user_password="" AND ms.system_password=""`
+		}
```
jonbarrow: I may be missing something, but why was this parameter called `matchmake`? I feel like something like `publicSession` or something would make more sense here, given that this filters out password protected sessions?

matchmake-extension/database/find_matchmake_session_by_search_criteria.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 // FindMatchmakeSessionBySearchCriteria finds matchmake sessions with the given search criterias
-func FindMatchmakeSessionBySearchCriteria(manager *common_globals.MatchmakingManager, connection *nex.PRUDPConnection, searchCriterias []match_making_types.MatchmakeSessionSearchCriteria, resultRange types.ResultRange, sourceMatchmakeSession *match_making_types.MatchmakeSession) ([]match_making_types.MatchmakeSession, *nex.Error) {
+func FindMatchmakeSessionBySearchCriteria(manager *common_globals.MatchmakingManager, connection *nex.PRUDPConnection, searchCriterias []match_making_types.MatchmakeSessionSearchCriteria, resultRange types.ResultRange, matchmake bool, sourceMatchmakeSession *match_making_types.MatchmakeSession) ([]match_making_types.MatchmakeSession, *nex.Error) {
```
jonbarrow: Not part of this review, but @DaniElectra some of these function signatures are getting pretty long. And now we're starting to introduce "magic values" to them with the `bool` here, which makes it unclear what the parameter is for at a glance

It might be time to consider moving to the config-struct model to clean some of this up a bit

matchmake-extension/join_matchmake_session_with_param.go
```diff
@@ -40,15 +40,19 @@ func (commonProtocol *CommonProtocol) joinMatchmakeSessionWithParam(err error, p
 		return nil, nexError
 	}
 
-	// TODO - Are these the correct error codes?
-	if bool(joinedMatchmakeSession.UserPasswordEnabled) && !joinMatchmakeSessionParam.StrUserPassword.Equals(joinedMatchmakeSession.UserPassword) {
+	// joinedMatchmakeSession.UserPasswordEnabled and joinedMatchmakeSession.SystemPasswordEnabled are only flags to inform the game wether the password is set,
+	// the passwords are always checked no matter what as can be seen by splatoon never setting the flags as it shares its info via friends instead of
+	// matchmaking lookup.
+	//
+	// if the password is unset and no password is specified they are empty anyhow so this wouldnt make a difference in the first place
```
jonbarrow: This comment feels largely unnecessary imo, it's clear by the changes themselves what is going on

jonbarrow (commented):
Not blocking anything rn, just leaving my 2 cents while waiting on @DaniElectra to also review. Thanks for this!

The only thing I might actually ask for changes on though is the parameter name and maybe updating the error messages, since the error codes and such are being touched anyway it might be a good time to update the relevant error messages

matchmake-extension/join_matchmake_session_with_param.go
RusticMaple: i removed the comment so hopefully that's fine now

matchmake-extension/database/find_matchmake_session_by_search_criteria.go
RusticMaple: i renamed the parameter to what you suggested

RusticMaple:
i adjusted the code with another commit to address the concerns(i dont know how this is handled here but if you want to keep a clean commit history i can squash both commits into one)

binaryoverload:
> i adjusted the code with another commit to address the concerns(i dont know how this is handled here but if you want to keep a clean commit history i can squash both commits into one)

Once a PR has been reviewed, it's good practice to not squash/rebase/etc - so this is good :)

RusticMaple:
oki thx for the feedback on that(its just that this happend to me once when i tried to contribute to the rust compiler that they told me to do that hence why i asked)

jonbarrow:
I appreciate you asking just in case, thank you

Changes LGTM, so I'll approve. @DaniElectra more-or-less owns the MM side of things though so I'll let him also review. Thanks again, we appreciate it!

jonbarrow (approved):

mrjvs:
@DaniElectra 👀

matchmake-extension/database/find_matchmake_session_by_search_criteria.go
DaniElectra: I agree, we could add some struct for this function (somthing like `FindMatchmakeSessionBySearchCriteriaOptions`? though that name feels very long) where we store these kind of parameters

DaniElectra (approved):
LGTM, thanks!

PR #66, opened 2026-07-23
feat: use options struct for FindMatchmakeSessionBySearchCriteria

DaniElectra:
Resolves #XXX

### Changes:

This is a proposal coming from #65 to simplify and clarify function parameters for our internal database functions, in this case for `FindMatchmakeSessionBySearchCriteria`. I'm not fully convinced of the options struct name `FindMatchmakeSessionBySearchCriteriaOptions` as it is very long, so other ideas are welcome

### DIFF
--- a/matchmake-extension/auto_matchmake_with_param_postpone.go
+++ b/matchmake-extension/auto_matchmake_with_param_postpone.go
@@ -48,7 +48,11 @@ func (commonProtocol *CommonProtocol) autoMatchmakeWithParamPostpone(err error, 
 
 	resultRange := types.NewResultRange()
 	resultRange.Length = 1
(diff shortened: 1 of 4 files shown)


jonbarrow:
I'm not home right now to do a proper review, but a couple things:

1. I think the name is also pretty long. Unfortunately we sometimes can't avoid that in Go and it *is* idiomatic, but we can probably shorten it to just like `FindMatchmakeSessionOptions` or something
2. Building off the first thing, renaming it to something shorter would let us get more generic with it. We could add more fields to this struct than what's currently being used and reuse it in different contexts, ignoring the fields in the consumer that we don't need and such
3. The options struct model usually means replacing all the function parameters with the struct, so in this case stuff like the manager and all that should be passed into the struct too

PR #68, opened 2026-07-26
Treat GRPC token errors as expired tokens

mrjvs:
Changes:
- Treat GRPC token errors as expired tokens

This resolves the issue where consoles have to be restarted every hour (tokens expire after an hour)

Every expired token enters the flow as a GRPC error ([see here](https://github.com/PretendoNetwork/account/blob/fc1206184f25e3703af948c4a0fdcf1b15f26ce4/src/services/grpc/account/v2/exchange-nex-token-for-user-data.ts#L14-L35)).
This was handled as a `ValidationFailed` NEX error. The console just gives that error back as a popup and kicks the user out of the game.

If we handle it as a `TokenExpired`, it tells the console to get a new token and try connecting again. This resolves all issues regarding expired tokens that we've had since the opaque tokens PR.

### DIFF
--- a/globals/authentication.go
+++ b/globals/authentication.go
@@ -69,7 +69,9 @@ func ValidatePNLoginData(pid types.PID, loginData types.DataHolder, gameServerID
 		Token:         tokenBase64,
 	})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


DaniElectra (approved):
Tested successfullt with friends. lgbt

PR #69, opened 2026-07-28
feat: storage manager and ranking legacy

ashquarky:
Resolves #XXX

### Changes:

Implements the old Storage Manager and Legacy Ranking protocols.
Notable things:
- Card IDs are just random and not actually saved on creation. This is to allow NN save files to carry over without exploding, as this implementation can adopt them on first use.
- Guessed implementation for unknown Storage Manager method 3 is included. Disabled since I don't have a game which uses it.
- All the Postgres queries are atomic one-shot things. This should make it really amenable to prepared statements and other optimisations.
- The unknown Ranking Legacy method 0xD is implemented based on Puyo Puyo 20th's observed behaviour. It displays the return value as the user's rank ("order" in protocol-speak) on a few screens, so I've called it "GetSelfRankingOrder". If we like this I can put it in nex-protocols-go and the wiki as well.
- For GetScore, half the RankingMode options seem impossible to implement, since they talk about "around self", but without a UniqueID, we don't know which "self" to use. Remember users can have multiple scores in the same category by using different UniqueIDs. Likely the RankingMode enum has different semantics on Legacy, but the impossible modes are left unimplemented for now and just give global results.
- The Limit parameter is unknown and discarded. Adding it may require leaderboard resets depending on what it does.

Still to be implemented:
- Postgres perf optimisations 
- NEX 1 CommonData allocation behaviour
- See README for more
- All the ResultCodes need to be audited and checked for correctness. Don't currently know what they *mean*, but the Nintendo Wiki provides example return codes.

Draft submission for initial prep and feedback, I know about the obvious things (LICENSE yay) needing polish but this is just a general look at it.
Currently in production for Puyo Puyo 20th and working well.


BDMCGaming:
I want to add a note that this currently is not working inside of Tetris Axis (NEX 2.0.0). You are able to login in once, but subsequent logins will result in [Core::SystemError] sql: no rows in result set during AcquireNexUniqueId.

PR #74, opened 2026-09-01
feat(match-making): implement GetDetailedParticipants

### LINKED ISSUE #72: [Enhancement]: Store strMessage from the Matchmaking functions in a table
### Checked Existing

- [x] I have checked the repository for duplicate issues.

### What enhancement would you like to see?

Currently, the strMessage inside of all the different matchmake functions (JoinMatchmakeSession, CreateCommunity, JoinCommunity, etc,.) gets discarded inside of the function and never touched aside from checking length.

It does seem that GetDetailedParticipants calls strMessage back through the ParticipantDetails type. Currently, it is not possible to implement this method without returning a false message. It would be ideal to be return the proper message rather than a stub though.

I observed this method being used inside MH3U, and it is presumably used by other games as well.

BDMCGaming:
Resolves #72 as a side effect

### Changes:

This will implement GetDetailedParticipants from MatchMakingProtocol. In the process, it resolves the issue mentioned above by creating a table that strMessage can be stored in with the gathering ID and user PID as the primary key. The table can then be queried any time strMessage is needed, such as by GetDetailedParticipants.

These changes have been tested and confirmed working with Monster Hunter 3 Ultimate. I'm going to go ahead and mark this as not an approved issue as implementing the method was not technically in the scope of the issue, though it is something I wish to see merged.

I'm more than happy to create a second pull request and/or issue to implement the method to split this one if that would be preferred.

### DIFF
--- a/globals/matchmake_referee_globals.go
+++ b/globals/matchmake_referee_globals.go
@@ -1,0 +1,22 @@
+package common_globals
+
+import (
+	"database/sql"
+	"sync"
+
+	"github.com/PretendoNetwork/nex-go/v2"
+)
+
+type MatchmakeRefereeManager struct {
+    Database *sql.DB
+    Endpoint *nex.PRUDPEndPoint
+    Mutex    *sync.RWMutex
+}
+
+func NewMatchmakeRefereeManager(endpoint *nex.PRUDPEndPoint, db *sql.DB) *MatchmakeRefereeManager {
+	return &MatchmakeRefereeManager{
+		Endpoint: endpoint,
+		Database: db,
+		Mutex:    &sync.RWMutex{},
+	}
+}
--- a/globals/utils.go
+++ b/globals/utils.go
@@ -57,7 +57,7 @@ func CheckValidMatchmakeSession(matchmakeSession match_making_types.MatchmakeSes
 		return false
 	}
 
-	if len(matchmakeSession.Attributes) != 6 {
+	if len(matchmakeSession.Attributes) != 9 {
 		return false
 	}
 
@@ -74,8 +74,8 @@ func CheckValidMatchmakeSession(matchmakeSession match_making_types.MatchmakeSes
 		return false
 	}
 
-	// * All buffers must have a length lower than 512
-	if len(matchmakeSession.ApplicationBuffer) > 512 {
+	// * All buffers must have a length lower than 1024
+	if len(matchmakeSession.ApplicationBuffer) > 1024 {
 		return false
 	}
 
--- a/match-making/database/record_message.go
+++ b/match-making/database/record_message.go
@@ -1,0 +1,16 @@
+package database
+
+import (
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+)
+
+func RecordMessage(manager *common_globals.MatchmakingManager, gatheringID uint32, PID types.PID, message string) *nex.Error {
+	_, err := manager.Database.Exec(`INSERT INTO matchmaking.messages (gathering_id, pid, message) VALUES ($1, $2, $3)`, gatheringID, PID, message)
+	if err != nil {
+		return nex.NewError(nex.ResultCodes.Core.SystemError, err.Error())
+	}
+
+	return nil
+}
--- a/match-making/get_detailed_participants.go
+++ b/match-making/get_detailed_participants.go
@@ -1,0 +1,38 @@
+package matchmaking
+
+import (
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+	"github.com/PretendoNetwork/nex-protocols-common-go/v2/match-making/database"
+	match_making "github.com/PretendoNetwork/nex-protocols-go/v2/match-making"
+)
+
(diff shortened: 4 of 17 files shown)


BDMCGaming:
I missed an outdated SQL column name from my previous/personal iteration of the method. This has now been corrected. I apologize for the mistake I made.

jonbarrow:
I don't believe this fully resolves #72. IIRC the `joinMessage` in classes like `AutoMatchmakeParam` and similar acts the same way as `strMessage` in this regard (since there's no other way for methods like `AutoMatchmakeWithParam_Postpone` to set this value on callers, or for methods like `JoinMatchmakeSessionWithParam` to se the value on the players being added via `additionalParticipants`)

I'm not sure that's within the scope of this PR, but it's something to keep in mind

globals/matchmake_referee_globals.go
```diff
@@ -0,0 +1,22 @@
...
+	"sync"
+
+	"github.com/PretendoNetwork/nex-go/v2"
+)
+
+type MatchmakeRefereeManager struct {
+    Database *sql.DB
+    Endpoint *nex.PRUDPEndPoint
+    Mutex    *sync.RWMutex
+}
```
jonbarrow, line 14: Just a nit: spaces are being used here instead of tabs. Go ships with a built in formatter, you should run that to catch everything https://go.dev/blog/gofmt

globals/utils.go
```diff
@@ -57,7 +57,7 @@ func CheckValidMatchmakeSession(matchmakeSession match_making_types.MatchmakeSes
 		return false
 	}
 
-	if len(matchmakeSession.Attributes) != 6 {
+	if len(matchmakeSession.Attributes) != 9 {
```
jonbarrow, line 60: This is incorrect and shouldn't be within the scope of this PR tbh. This value is derived from the real constant [`NUM_MATCHMAKE_SESSION_ATTRIBUTES `](https://github.com/PretendoNetwork/nex-protocols-go/blob/a2ae8b73a00876653ad50b1a19a49d8a4581ea10/match-making/constants/constants.go#L10-L11). The 9 value is a game-specific modification

globals/utils.go
```diff
@@ -74,8 +74,8 @@ func CheckValidMatchmakeSession(matchmakeSession match_making_types.MatchmakeSes
 		return false
 	}
 
-	// * All buffers must have a length lower than 512
-	if len(matchmakeSession.ApplicationBuffer) > 512 {
+	// * All buffers must have a length lower than 1024
+	if len(matchmakeSession.ApplicationBuffer) > 1024 {
```
jonbarrow, line 78: This is incorrect and shouldn't be within the scope of this PR tbh. This value is derived from the real constant [`MATCHMAKE_BUFFER_MAX_LENGTH `](https://github.com/PretendoNetwork/nex-protocols-go/blob/a2ae8b73a00876653ad50b1a19a49d8a4581ea10/match-making/constants/constants.go#L84-L85). This a game-specific modification

match-making/database/record_message.go
```diff
@@ -0,0 +1,16 @@
+package database
+
+import (
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+)
+
+func RecordMessage(manager *common_globals.MatchmakingManager, gatheringID uint32, PID types.PID, message string) *nex.Error {
```
jonbarrow, line 9: Exported functions should have a Godoc comment, but this doesn't seem to get used anywhere outside of `database` so it can be unexported as well (though a Godoc comment would still be nice either way for docs purposes)

BDMCGaming:
Actually, I did go into this with those in mind. I placed the RecordMessage call inside of JoinGathering and JoinGatheringWithParticipants as that is the base layer for joining. JoinMatchmakeSession and its variants all call one of the two functions inside with it, and the AutoMatchmake handlers already call JoinMatchmakeSession. CreateCommunity also calls  JoinGathering.

From what I know, all the different entry points are accounted for.

match-making/get_detailed_participants.go
```diff
@@ -0,0 +1,38 @@
...
+
+	rmcResponseBody := rmcResponseStream.Bytes()
+
+	rmcResponse := nex.NewRMCSuccess(endpoint, rmcResponseBody)
+	rmcResponse.ProtocolID = match_making.ProtocolID
+	rmcResponse.MethodID = match_making.MethodGetDetailedParticipants
+	rmcResponse.CallID = callID
+
+	return rmcResponse, nil
+}
```
jonbarrow, line 38: Until we see a game try to call this method without being in the target gathering, I think this should be limited to only letting players already in the gathering request this data. Just so someone can't spam the API to look for gatherings a specific person is in

jonbarrow (commented):
I stopped reviewing after the `GetDetailedParticipants` parts, this PR seems to add many more methods and makes several other changes out of the defined scope. This touches several protocols in ways that I don't believe are related and should probably be their own [PRs to keep scope limited](https://github.com/PretendoNetwork/.github/blob/master/.github/CONTRIBUTING.md#pull-requests)

BDMCGaming:
Okay I'm just going to remake the PR. I'm so sorry. I have no idea how those files leaked into this pull request, as I don't recall staging them inside of Visual Studio.

BDMCGaming:
I'll get to work on resolving the issue you mentioned in the handler tomorrow, as it is getting late for me.

For reference, the util.go are personal changes that I've made so I can temporarily get past the blocker until it is properly resolved. The Matchmake_Referee_globals.go was part of another project I was working on that got halted due to a bug in MatchmakeExtension.

jonbarrow:
All good, not the end of the world. It happens 👍

PR #75, opened 2026-09-01
feat(match-making): implement GetDetailedParticipants

### LINKED ISSUE #72: [Enhancement]: Store strMessage from the Matchmaking functions in a table

BDMCGaming:
Let's try this again, preferably without file leakage.

Resolves #72 as a side effect.

### Changes:

This will implement GetDetailedParticipants from MatchMakingProtocol. In the process, it resolves the issue mentioned above by creating a table that strMessage can be stored in with the gathering ID and user PID as the primary key. The table can then be queried any time strMessage is needed, such as by GetDetailedParticipants. 

Database function RecordMessage has been inserted in Database functions JoinGathering and JoinGatheringWithParticipants to account for this. All other matchmaking functions call one of the variants while it is joining a MatchmakeSession, PersistentGathering, whatever. This makes those two functions the most optimal place to execute the functions as it will all lead to JoinGathering eventually.

These changes have been tested and confirmed working with JoinMatchmakeSessionEx and one of the AutoMatchmake methods. These were tested with Monster Hunter 3 Ultimate and Kirby Battle Royale respectively, and both tests were successful. I am going to mark this as not an approved issue due to GetDetailedParticipants technically being outside the scope of the original issue (though it was a requirement for this method to be implemented regardless).

### DIFF
--- a/match-making/database/get_detailed_participants.go
+++ b/match-making/database/get_detailed_participants.go
@@ -1,0 +1,60 @@
+package database
+
+import (
+	"database/sql"
+
+	"github.com/PretendoNetwork/pq-extended"
+	"slices"
+
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+	matchmaking_types "github.com/PretendoNetwork/nex-protocols-go/v2/match-making/types"
+)
+
+// GetDetailedParticipants takes an array of participants associated with a gathering ID and creates info for each participant
+func GetDetailedParticipants(manager *common_globals.MatchmakingManager, gatheringID types.UInt32, sourcePID types.PID) (types.List[matchmaking_types.ParticipantDetails], *nex.Error) {
+	var participantList []uint64
+
+	err := manager.Database.QueryRow(`SELECT participants FROM matchmaking.gatherings WHERE id = $1`, gatheringID).Scan(pqextended.Array(&participantList))
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return types.NewList[matchmaking_types.ParticipantDetails](), nex.NewError(nex.ResultCodes.RendezVous.SessionVoid, err.Error())
+		} else {
+			return types.NewList[matchmaking_types.ParticipantDetails](), nex.NewError(nex.ResultCodes.Core.Unknown, err.Error())
+		}
+	}
+
+	if len(participantList) == 0 {
+		// * Empty gathering
+		return types.NewList[matchmaking_types.ParticipantDetails](), nil
+	}
+
+	if !slices.Contains(participantList, uint64(sourcePID)) {
+		// * Is this the right error to use?
+		return types.NewList[matchmaking_types.ParticipantDetails](), nex.NewError(nex.ResultCodes.RendezVous.NotParticipatedGathering, "User attempted GetDetailedParticipants inside of a NotParticipatedGathering")
+	}
+
+	var participantDetails types.List[matchmaking_types.ParticipantDetails]
+
+	for _, participant := range participantList {
+		participantInfo := matchmaking_types.NewParticipantDetails()
+		err = manager.Database.QueryRow(`SELECT pid, message FROM matchmaking.messages WHERE gathering_id = $1 AND pid = $2`, gatheringID, participant).Scan(&participantInfo.IDParticipant, &participantInfo.StrMessage)
+		if err != nil {
+			common_globals.Logger.Error(err.Error())
+			continue
+		}
+
+		accountDetails, err := manager.Endpoint.AccountDetailsByPID(types.NewPID(participant))
+		if err != nil {
+			return types.NewList[matchmaking_types.ParticipantDetails](), nex.NewError(nex.ResultCodes.Core.Unknown, err.Error())
+		}
+
+		participantInfo.StrName = types.String(accountDetails.Username)
+		participantInfo.UIParticipants = types.UInt16(len(participantList))
+
+		participantDetails = append(participantDetails, participantInfo)
+	}
+
+	return participantDetails, nil
+}
--- a/match-making/database/get_detailed_participants.go
+++ b/match-making/database/get_detailed_participants.go
@@ -12,7 +12,7 @@ import (
 	matchmaking_types "github.com/PretendoNetwork/nex-protocols-go/v2/match-making/types"
 )
 
-// GetDetailedParticipants takes an array of participants associated with a gathering ID and creates info for each participant
+// GetDetailedParticipants returns detailed information about the participants of the given gathering
 func GetDetailedParticipants(manager *common_globals.MatchmakingManager, gatheringID types.UInt32, sourcePID types.PID) (types.List[matchmaking_types.ParticipantDetails], *nex.Error) {
 	var participantList []uint64
 
@@ -37,9 +37,16 @@ func GetDetailedParticipants(manager *common_globals.MatchmakingManager, gatheri
 
 	var participantDetails types.List[matchmaking_types.ParticipantDetails]
 
-	for _, participant := range participantList {
+	for _, participant := range common_globals.RemoveDuplicates(participantList) {
 		participantInfo := matchmaking_types.NewParticipantDetails()
-		err = manager.Database.QueryRow(`SELECT pid, message FROM matchmaking.messages WHERE gathering_id = $1 AND pid = $2`, gatheringID, participant).Scan(&participantInfo.IDParticipant, &participantInfo.StrMessage)
+
+		for _, pid := range participantList {
+			if participant == pid {
+				participantInfo.UIParticipants += 1
+			}
+		}
+
+		err = manager.Database.QueryRow(`SELECT pid, message FROM matchmaking.join_messages WHERE gathering_id = $1 AND pid = $2`, gatheringID, participant).Scan(&participantInfo.IDParticipant, &participantInfo.StrMessage)
 		if err != nil {
 			common_globals.Logger.Error(err.Error())
 			continue
@@ -47,11 +54,11 @@ func GetDetailedParticipants(manager *common_globals.MatchmakingManager, gatheri
 
 		accountDetails, err := manager.Endpoint.AccountDetailsByPID(types.NewPID(participant))
 		if err != nil {
-			return types.NewList[matchmaking_types.ParticipantDetails](), nex.NewError(nex.ResultCodes.Core.Unknown, err.Error())
+			common_globals.Logger.Error(err.Error())
+			continue
 		}
 
 		participantInfo.StrName = types.String(accountDetails.Username)
-		participantInfo.UIParticipants = types.UInt16(len(participantList))
 
 		participantDetails = append(participantDetails, participantInfo)
 	}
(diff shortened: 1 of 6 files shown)


BDMCGaming:
I did also use Jon's feedback in the previous pull request on implementing a check for the user being in the gathering that. In this case, the client will be returned with RendezVous::NotParticipatedGathering with a message being placed in logs that a user attempted to call the method on a gathering they were not in.

BDMCGaming:
@jonbarrow May you please review this whenever you have the free time? I believe I have the pull request done properly this time.

BDMCGaming:
I wanted to bump this, as I'm unsure if it got missed or if something happened.

jonbarrow:
It hasn't been missed. I've just been very sick recently and not able to focus on real reviews right now. I'll tag @DaniElectra though

match-making/protocol.go
```diff
@@ -56,6 +56,17 @@ func (commonProtocol *CommonProtocol) SetManager(manager *common_globals.Matchma
 		return
 	}
 
+	_, err = manager.Database.Exec(`CREATE TABLE IF NOT EXISTS matchmaking.messages (
```
DaniElectra: This should probably be named differently, as there can be messages for both joining and leaving. Maybe `join_messages` or something like that

BDMCGaming:
Ah. I apologize for bringing it up then, Jon.

match-making/database/get_detailed_participants.go
```diff
@@ -0,0 +1,60 @@
...
+			continue
+		}
+
+		accountDetails, err := manager.Endpoint.AccountDetailsByPID(types.NewPID(participant))
+		if err != nil {
+			return types.NewList[matchmaking_types.ParticipantDetails](), nex.NewError(nex.ResultCodes.Core.Unknown, err.Error())
+		}
+
+		participantInfo.StrName = types.String(accountDetails.Username)
+		participantInfo.UIParticipants = types.UInt16(len(participantList))
```
DaniElectra: Pretty sure this represents the number of participants that have joined *under* this PID. Not to be confused with the whole "Added to gathering" notification stuff, there can be multiple participants playing from one same PID (like having multiple people with different controllers playing online from the same console). Think of it like "guest accounts" that are attached to one main PID. This is real functionality that is used in games like MK8 for example

The matchmaking implementation accounts for this by having duplicated the PID as per the number of participants that are attached to it, so the proper way to get this value would be to count the number of times that the PID appears on the participant list. Keep in mind that you will also need to remove the duplicate PIDs from the list before returning it to the RMC! You can do that with the helper function `globals.RemoveDuplicates`

match-making/database/get_detailed_participants.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/PretendoNetwork/pq-extended"
+	"slices"
+
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+	matchmaking_types "github.com/PretendoNetwork/nex-protocols-go/v2/match-making/types"
+)
+
+// GetDetailedParticipants takes an array of participants associated with a gathering ID and creates info for each participant
```
DaniElectra:
```suggestion
// GetDetailedParticipants returns detailed information about the participants of the given gathering
```

match-making/get_detailed_participants.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+	"github.com/PretendoNetwork/nex-protocols-common-go/v2/match-making/database"
+	match_making "github.com/PretendoNetwork/nex-protocols-go/v2/match-making"
+)
+
+func (commonProtocol *CommonProtocol) GetDetailedParticipants(err error, packet nex.PacketInterface, callID uint32, idGathering types.UInt32) (*nex.RMCMessage, *nex.Error) {
```
DaniElectra:
```suggestion
func (commonProtocol *CommonProtocol) getDetailedParticipants(err error, packet nex.PacketInterface, callID uint32, idGathering types.UInt32) (*nex.RMCMessage, *nex.Error) {
```

For consistency with other functions at this level

match-making/database/record_message.go
```diff
@@ -0,0 +1,17 @@
+package database
+
+import (
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+)
+
+// This function records the strMessage or joinMessage from a Match-Making RMC call and stores it in a table for future queries
+func RecordMessage(manager *common_globals.MatchmakingManager, gatheringID uint32, PID types.PID, message string) *nex.Error {
```
DaniElectra:
```suggestion
// RecordJoinMessage records the join message from a participant into a gathering
func RecordJoinMessage(manager *common_globals.MatchmakingManager, gatheringID uint32, PID types.PID, message string) *nex.Error {
```

This might be better

match-making/database/record_message.go
```diff
@@ -0,0 +1,17 @@
...
+	_, err := manager.Database.Exec(`INSERT INTO matchmaking.messages (gathering_id, pid, message) VALUES ($1, $2, $3)`, gatheringID, PID, message)
```
DaniElectra: A user can leave and join the same gathering multiple times, so the SQL query must account for that too. I think doing an `ON CONFLICT DO UPDATE ...` updating the message should be good enough

DaniElectra (changes requested):
Thanks for your contribution! There are various details that need to be accounted for that aren't exactly well known, but otherwise it's looking good!

match-making/protocol.go
BDMCGaming: If this is the case, should I create a table for leave messages as well? I'm unsure if those ever get used by any methods, but it may be beneficial to store those if they do get used.

match-making/database/get_detailed_participants.go
BDMCGaming: I was told by Jon in the one issue I made that it is just the number of users in a gathering. It would be nice if I had a dump of this method, but I don't believe any of my Switch games call this method.

I have conflicting information now, so I'm not too sure which way I should go with.

match-making/database/get_detailed_participants.go
BDMCGaming: Sorry! I've never been a good docs person unfortunately, especially with things like these. This does sound much better than how I explained it!

match-making/database/get_detailed_participants.go
jonbarrow: > I was told by Jon in the one issue I made that it is just the number of users in a gathering

It's entirely possible I'm wrong/misunderstood what I was looking at, when it comes to matchmaking Dani does generally know more than I do in places. I would generally take his word over mine in these cases, unless I'm adamant about something

BDMCGaming: Alright. In this case, I'll go with this suggestion then.

BDMCGaming: Okay. I'm sorry, but I'm not too sure how I would go about incrementing this value if there is a duplicate PID/2nd user. May you please tell me how this would be possible?

BDMCGaming:
I've got all of these changes made on my end except for the change to UIParticipants. I'm unsure how to properly handle the duplicates at the moment.

If I run RemoveDuplicates, I know that the player count will end up being inaccurate as a player is missing. I'm not sure how to increment the UIParticipants for duplicates though.

DaniElectra:
You can do something like this:

```go
	for _, participant := range globals.RemoveDuplicates(participantList) {
		participantInfo := matchmaking_types.NewParticipantDetails()

		// ...

		participantInfo.StrName = types.String(accountDetails.Username)

		for _, pid := range participantList {
			if participant == pid {
				participantInfo.UIParticipants += 1
			}
		}

		participantDetails = append(participantDetails, participantInfo)
	}
```

BDMCGaming:
I went ahead and included details of what got changed inside of the commit description. I also made a small adjustment to error handling in the for/range loop, as I realized that a non-existent PID will completely halt the method during GRPC rather than just skipping the participant (there is a single game this applies to lmao).

BDMCGaming:
I did not get a chance to test the UIParticipants change unfortunately, as the only game I have that uses it predates the parameter. Unless one of the devs have a game that uses it, all that can really be done is hope it works.

BDMCGaming:
Hello?

BDMCGaming:
May I please ask what is going on? I have not heard back in almost a week now, so I'm unsure if something is wrong with the pull request or what.

jonbarrow (dismissed):
LGTM, anything I missed @DaniElectra?

match-making/database/record_join_message.go
DaniElectra, line 1: The file needs to be renamed as per the function name to `record_join_message.go`

DaniElectra (commented):
only one small thing

BDMCGaming:
Apologies for that. It should be fixed now.

match-making/database/record_join_message.go
```diff
@@ -0,0 +1,17 @@
+package database
+
+import (
+	"github.com/PretendoNetwork/nex-go/v2"
+	"github.com/PretendoNetwork/nex-go/v2/types"
+	common_globals "github.com/PretendoNetwork/nex-protocols-common-go/v2/globals"
+)
+
+// RecordJoinMessage records the join message from a participant into a gathering
+func RecordJoinMessage(manager *common_globals.MatchmakingManager, gatheringID uint32, PID types.PID, message string) *nex.Error {
```
DaniElectra:
```suggestion
func RecordJoinMessage(manager *common_globals.MatchmakingManager, gatheringID uint32, userPID types.PID, message string) *nex.Error {
```

Sorry just noticed this. It generally isn't good practice to use PascalCase on variable names that aren't constants, camelCase is the right choice here. Altermatively you can do just lowercase `pid`

BDMCGaming: I'll apply it, but do you mind if I ask why?

DaniElectra: Go is special in that the name of a variable / constant / function / method can affect who can access it. Those who start with uppercase are public, while staring in lowercase means it's private. Using local variables in functions with capitalized names can create confusion as to whether this is a local variable or a constant defined within the same package

DaniElectra (approved):
LGTM, thanks!

match-making/database/record_join_message.go
BDMCGaming: Oh my god now it makes sense why Go keeps saying it cannot access a function I have imported. Thank you so much for telling me about this Dani.

PR #77, opened 2026-09-20
fix: fix community participation count

BDMCGaming:
Resolves #XXX

While I was looking into Monster Hunter 3 Ultimate, I found that m_ParticipationCount was not being implemented correctly. After some digging with Jon inside of #dev-supporters, we found that the current implementation only is accurate to (what seems to be) a single game.

We found through some Mario Golf dumps that this parameter is supposed to actually be a count of everybody inside of the Persistent Gathering. This does line up with how I see it used in Monster Hunter 3 Ultimate, and presumably other NEX games.

# Changes

This pull requests make the following changes to correct this behavior and accomodate for the correct behavior:

- The participation_count column has been removed from the persistent gathering table.

- UpdatePersistentGatheringParticipationCount is split into two functions that account for a user joining and leaving a persistent gathering
  - AddPersistentGatheringParticipation is intended to be used when joining the community (through CreateCommunity or JoinCommunity)
  - RemovePersistentGatheringParticipation is intended to be used when leaving the community (through EndParticipation)
- All database functions that fetch Persistent Gatherings are tweaked to count the rows associated with a PID rather than retrieving the count associated with a (user_pid, gathering_id) combo
- UpdatePersistentGatheringParticipation is removed from JoinMatchmakeSession and JoinMatchmakeSessionWithParticipants

All current changes have been tested with Monster Hunter 3 Ultimate, and I can do additional testing with Tetris Ultimate if needed.


BDMCGaming:
I'm keeping this as a draft as I try to decide the best way to implement Add/RemovePersistentGatheringParticipation into the existing methods without creating an accidental import loop.

I believe I'll drop RemovePersistentGatheringParticipation into the handler of EndParticipation since the gathering type is already exposed in there, and AddPersistentGatheringParticipation will likely go inside of CreateCommunity and JoinCommunity (once implemented) directly.

matchmake-extension/database/join_matchmake_session_with_participants.go
```diff
@@ -17,31 +16,5 @@ func JoinMatchmakeSessionWithParticipants(manager *common_globals.MatchmakingMan
 		return 0, nexError
 	}
 
-	// TODO - Should we return the error in these cases?
-	if uint32(matchmakeSession.MatchmakeSystemType) == uint32(constants.MatchmakeSystemTypePersistentGathering) { // * Attached to a persistent gathering
```
TraceEntertains, line 21: Should this be block removed completely (same question applies to database/JoinMatchmakeSession)? I believe it might be nice to keep the LogParticipateCommunity call for moderator tooling

matchmake-extension/protocol.go
```diff
@@ -97,7 +97,6 @@ func (commonProtocol *CommonProtocol) SetManager(manager *common_globals.Matchma
 		id bigserial PRIMARY KEY,
 		user_pid numeric(20),
 		gathering_id bigint,
-		participation_count bigint,
```
TraceEntertains, line 100: I did mention this in the discord earlier, but I feel it would be best to bring up here as well, I think it would be worth having a discussion for if this table should be removed entirely; with these changes, it serves little purpose other than being able to query for PersistentGathering IDs a user is in directly (rather than querying for PersistentGatherings with that user in the participants), and that is a relatively niche and somewhat insignificant use-case to have a whole extra table for in my opinion, especially considering that, to my knowledge, we don't do anything like that in other places unless it's deemed absolutely necessary

TraceEntertains (commented):
Didn't really have much to comment on stylistically (or in general really), just wanted to mention a couple things implementation wise that I think should be discussed further and which I didn't want to see get glossed over

matchmake-extension/database/join_matchmake_session_with_participants.go
BDMCGaming: I'm working off the assumption that LogParticipateCommunity is for when a user joins a community. Right now, this is calling it everytime a user joins a Matchmake Session associated with a persistent gathering. If this is wrong from a moderation stand point, please tell me so.

matchmake-extension/protocol.go
BDMCGaming: Actually, this isn't how I intend for it to work.

Jon said that users can only participate in so many PersistentGatherings at a time. At the time being, this table is unused aside from populating m_ParticipationCount. It is needed for JoinCommunity once that is implemented, as I need to be able to check for the amount of PersistentGatherings the user is in.

BDMCGaming:
> I believe I'll drop RemovePersistentGatheringParticipation into the handler of EndParticipation since the gathering type is already exposed in there...

I misremembered this, and there is a small dilemma as a result of this. The gathering type is exposed inside of EndGatheringParticipation (which is called in the handler). I can't place RemovePersistentGatheringparticipation inside of that function as then we would have an import loop going on.

I presume I could change the output of the function to additionally send the gathering type with the nexError, though I would prefer to ask if this is an alright change before implementing it.

matchmake-extension/database/join_matchmake_session_with_participants.go
TraceEntertains: Ahhh, gotcha
To be fair I haven't really looked to deep into PersistentGathering-adjacent matchmaking code for a while, so that's probably fine then lol
Marking as resolved

matchmake-extension/protocol.go
TraceEntertains: To be fair, you can also find that through querying the gatherings table directly

BDMCGaming:
@DaniElectra May you please review this pull request and review the above I brought up about EndGatheringParticipation?

BDMCGaming:
Thought about this a bit more. Would it be beneficial to move tracking.LogParticipateCommunity inside of AddPersistentGatheringParticipation rather adding it into JoinCommunity as I originally planned?

BDMCGaming:
Looking into this more as I troubleshoot/deal with other things. I believe it would needed to place EndPersistentGatheringParticipation inside of the DisconnectParticipant database function as well since this is another way a user can be removed from a persistent gathering. If this is actually needed, then it brings up the problem of an import loop.

Are these functions better off being inside of the matchmaking database so that this import loop doesn't need to happen, or is there another way this problem can be approached that allows?

